Issue #1, opened 2018-10-29
Please add a TypeScript definition file


kristianmandrup:
Perhaps to start with, something as simple as:

```ts
declare module "react-apollo-hooks" {
  function ApolloProvider({ children, client }: any): any;
  function useApolloClient(): any;
  function useApolloQuery(query: any, { variables }?: any): any;
  function useApolloMutation(mutation: any, baseOptions?: any): any;
}
```

cdebotton:
@kristianmandrup Here ya go:
https://github.com/trojanowski/react-apollo-hooks/pull/2

trojanowski:
Hi @kristianmandrup. What do you think about #2 and #3?

kristianmandrup:
Hi @trojanski, looks very good!! Amazing job! Thanks :)

PR #2, opened 2018-10-30
Add typscript definitions

cdebotton:
Addresses #1 

This should follow the type definitions for the query and mutation components relatively closely.

Example:

```tsx
// With useApolloQuery

function AdminUsers(props: { className?: string } & RouteComponentProps<void>) {
  const { className, match } = props;
  const {
    data: { users, session },
  } = useApolloQuery<Users>(USERS_QUERY);

  return (
    <InsetLayout className={className}>
      <UsersHeading>Users</UsersHeading>
      <NewUserLink to={`${match.url}/new`}>New user</NewUserLink>

      <UserList>
        {users.map(user => {
          const isCurrentUser =
            (session && session.user.id === user.id) || false;
          return (
            <UserListItem
              key={user.id}
              user={user}
              baseUrl={match.url}
              isCurrentUser={isCurrentUser}
            />
          );
        })}
      </UserList>
    </InsetLayout>
  );
}
```

```tsx
// With useApolloMutation

function DeleteUserButton(props: {
  user: Users_users;
}) {
  const { user } = props;
  const mutate = useApolloMutation<
    DeleteUserMutation,
    DeleteUserMutationVariables
  >(DELETE_USER_MUTATION, {
    variables: { where: { id: user.id } },
  });

  return (
    <DeleteIcon
      onClick={mutate}
    >
      <Tooltip content={`Delete ${user.email}`}>
        <AnimatedCross />
      </Tooltip>
    </DeleteIcon>
  );
}
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,16 @@
+import { NormalizedCacheObject } from 'apollo-cache-inmemory';
+import ApolloClient, { ApolloQueryResult, OperationVariables } from 'apollo-client';
+import { DocumentNode } from 'graphql';
+import { ReactNode } from 'react';
+import { MutationFn, MutationOptions } from 'react-apollo';
+interface ProviderProps {
+    children: ReactNode;
+    client: ApolloClient<NormalizedCacheObject>;
+}
+export declare function ApolloProvider({ children, client }: ProviderProps): JSX.Element;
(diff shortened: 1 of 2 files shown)


trojanowski:
Hi @cdebotton. Thanks for your PR. I prepared an alternative one: #3. The main difference is that one doesn't depend on react-apollo. Looking forward to your comments.

Issue #6, opened 2018-11-04
Rename hooks


nicolaslopezj:
It is necessary to include the Apollo word in function names? I think it would be much cleaner without it

```useQuery```

```useMutation```

We could have both and deprecate the old names with a message.

trojanowski:
Hi. @nicolaslopezj Yes, I think it's a good idea. I'm going to implement it (or accept a PR).

PR #9, opened 2018-11-05
Remove the "Apollo" word from the api

nicolaslopezj:
Hooks renamed:

- `useApolloClient` -> `useClient`
- `useApolloQuery` -> `useQuery`
- `useApolloMutation` -> `useMutation`

Also, warning messages were created

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -19,13 +19,11 @@ export function ApolloProvider<CacheShape = any>(props: {
   client: ApolloClient<CacheShape>;
 }): JSX.Element;
 
-export function useApolloClient<CacheShape = any>(): ApolloClient<
-  CacheShape
-> | null;
+export function useClient<CacheShape = any>(): ApolloClient<CacheShape> | null;
 
 type QueryHookOptions<TVariables> = Omit<QueryOptions<TVariables>, 'query'>;
 
-export function useApolloQuery<TData = any, TVariables = OperationVariables>(
+export function useQuery<TData = any, TVariables = OperationVariables>(
   query: DocumentNode,
   options?: QueryHookOptions<TVariables>
 ): ApolloQueryResult<TData> & {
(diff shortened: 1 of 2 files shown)


nicolaslopezj:
This closes #6

trojanowski:
Hi @nicolaslopezj. Thanks for the PR. I've got a few suggestions.

1. I'd stay with `useApolloClient`. `useClient` seems to be too generic for me, especially as it's used less often than other function. Also react-apollo uses `Query` and `Mutation`, but `ApolloConsumer` and `ApolloProvider`.
2. Could you use a package like [warning](https://www.npmjs.com/package/warning) instead of `console.warning()` directly? It will ease locating of usage of the deprecated methods if they were imported under different names.
3. Could you leave the deprecated methods in typescript definition (marked as deprecated). Adding something like this in index.d.ts seems works:

```typescript
/**
 * @deprecated use useQuery instead
 */
export const useApolloQuery: typeof useQuery;
/**
 * @deprecated use useMutation instead
 */
export const useApolloMutation: typeof useMutation;
```

nicolaslopezj:
I think useClient is good, because it’s being imported from a package that has the Apollo work in it. The idea was to mimic the name of the high order component withClient provided by Apollo. 

Could you do N3? I don’t know how to write typescript

Thank you!

trojanowski:
> I think useClient is good, because it’s being imported from a package that has the Apollo work in it. The idea was to mimic the name of the high order component withClient provided by Apollo.

The HOC is actually `withApollo`. :) I'd say that `useApolloClient` is better than `useApollo` because is more consistent with render props react-apollo API (_Apollo_ is a part of render props components used for providing and consuming the client).

> Could you do N3? I don’t know how to write typescript

Sure, but it's actually enough to put the snippet from my previous comment at the bottom of _index.d.ts_.

trojanowski:
@nicolaslopezj 

I published a version 0.1.3 with `useQuery` and `useMutation` hooks.

PR #15, opened 2018-11-10
WIP: Allow to use 'cache-and-network' fetch policy with suspense

### LINKED ISSUE #13: Using fetchPolicy: 'cache-and-network' creates infinite loop
Using the options `cache-and-network` for fetch policy creates an infinite loop.

I could see that in this `if`:

```js
query === previousQuery.current &&
isEqual(variables, previousVariables.current) &&
isEqual(apolloContextOptions, previousApolloContextOptions.current) &&
isEqual(restOptions, previousRestOptions.current)
```

```previousVariables.current``` and ```previousRestOptions.current``` and always undefined.

I don't know why useRef is not persisting.

trojanowski:
Closes #13

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -33,12 +33,18 @@ export function useQuery(
   const previousApolloContextOptions = useRef();
   const previousRestOptions = useRef();
   const observableQuery = useRef();
+  const shouldRefetch = useRef();
 
   useEffect(
     () => {
       const subscription = observableQuery.current.subscribe(nextResult => {
         setResult(nextResult);
       });
+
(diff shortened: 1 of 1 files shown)


trojanowski:
It tries to solve #13 by emulating the cache-and-network policy by using the default `cache-first` one and then optionally forcing a refetch. The problem with that that the refetch is also invoked the first time the page is loading.

trojanowski:
I'm closing it - it's an outdated implementation. I'll try to create a new one when Suspense for data fetching is considered ready by the React team - maybe it'll be easier and less hacky then.

PR #27, opened 2018-11-29
Allow configure default value for suspend globally

langpavel:
Few lines of code better that thousand words :slightly_smiling_face:

Code is fully compatible but I will prefer suspense initially disabled.
Also `.d.ts` typings is missing, sorry about that, treat this as better issue/question :wink:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,6 +3,16 @@ import isEqual from 'react-fast-compare';
 
 import deprecated from './deprecated';
 
+const globalConfig = {
+  suspend: true, // should be false for now?
+};
+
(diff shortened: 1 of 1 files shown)


trojanowski:
Hi @langpavel 

Thanks for your PR, but I don't think that the global config is a good idea. If you want to use non-suspense mode by default you can very easily create a custom hook like this one:

```javascript
import { useQuery as useQueryBase } from 'react-apollo-hooks';

export function useQuery(query, options = {}) {
  return useQueryBase(query, { ...options, suspend: false });
}

```

langpavel:
Good point, thanks!

Issue #21, opened 2018-11-21
Conditionally firing a query


shlomokraus:
I am trying to implement a conditional pattern which I am not sure how it should work with hooks. 
I have one query asking for currentUser and only when and if it returns a userId I will execute the query. 

By the rules of hooks, you must not nest hooks inside conditions or loops, so how can I achieve this? 

This is a sample of my custom hook: 

```ts
function useSomething() { 
  const [userId] = useCurrentUser();
  // Here we need to skip the query
  const { data, errors } = useQuery(query, { variables: {userId} });
  // Do some stuff and return data
 return [data, errors]
}
```

trojanowski:
Hi @shlomokraus

Right now the only way to do it is to conditionally render a component using the `useQuery` hook. I think we could add `skip` option to it.  The query wouldn't run if it were equal to true.
Also please look at https://github.com/trojanowski/react-apollo-hooks/issues/18#issuecomment-438191372

shlomokraus:
I also think that adding "skip" or similar is the best solution. Rendering the component conditionally will break encapsulation.

PR #35, opened 2018-12-17
Forward the subscribeToMore option to enable usage from useQuery

dimitar-nikovski:


seeden:
I did not see subscribeToMore in the observer. Did you test it ? Is it working ?

dimitar-nikovski:
Hi @seeden , indeed it is part of the prototype of ObservableQuery from apollo-client/core [link](https://github.com/apollographql/apollo-client/blob/911ca2db631adc38745b8098dd802e0cdfe8dbbb/packages/apollo-client/src/core/ObservableQuery.ts#L359) and is available in the returned observableQuery.

The relevant excerpt of calls is [useQuery](https://github.com/trojanowski/react-apollo-hooks/blob/e1a9b6f77d2f9fb8c34ad89ee0ec5e3720e1b16d/src/index.js#L61) -> [getCachedObservableQuery](https://github.com/trojanowski/react-apollo-hooks/blob/e1a9b6f77d2f9fb8c34ad89ee0ec5e3720e1b16d/src/queryCache.js#L12) -> [watchQuery](https://github.com/apollographql/apollo-client/blob/911ca2db631adc38745b8098dd802e0cdfe8dbbb/packages/apollo-client/src/ApolloClient.ts#L259) -> instance of [ObservableQuery](https://github.com/apollographql/apollo-client/blob/911ca2db631adc38745b8098dd802e0cdfe8dbbb/packages/apollo-client/src/core/QueryManager.ts#L672) -> [public method subscribeToMore](https://github.com/apollographql/apollo-client/blob/911ca2db631adc38745b8098dd802e0cdfe8dbbb/packages/apollo-client/src/core/ObservableQuery.ts#L359)

seeden:
Great sounds good. I can not wait :)

trojanowski:
@dimitar-nikovski Thanks for your PR, but I'm not sure if it's a good idea to use `subscribeToMore` with hooks. You'll have to use it with the `useEffect` or `useLayoutEffect` hook and ensure that the array with correct variables is passed as the second argument of that hook and a proper unsubscribe function returned. Otherwise, you'll risk that the subscription will be subscribed/unsubscribed after each update (even if arguments won't be changed) or not resubscribed at all if they changed. In my opinion, a much better solution is the `useSubscription` hook which is WIP in #37. Are you aware of any use cases of `subscribeToMore` which wouldn't be possible to solve with `useSubscription` (or [ApolloClient.subscribe](https://www.apollographql.com/docs/react/api/apollo-client.html#ApolloClient.subscribe) which it will use under the hood)?

seeden:
I am using it like this

```js
export default (query, options: Options = {}): Object => {
  const { onReady, ...rest } = options;
  const [subscribed, setSubscribed] = useState(false);
  const response = useQuery(query, {
    suspend: false,
    ...rest,
  });

  const { loading, error } = response;
  if (error) {
    throw error;
  }

  if (!loading && !subscribed && onReady) {
    setSubscribed(true);
    onReady(response);
  }

  return response;
};
```


useQuery(query, {
  onReady({ subscribeToMore } => {
      ...
  }),
})

trojanowski:
@seeden What do you think about that API:

```javascript
const response = useQuery(query, {
  // query options
});

const subscriptionResult = useSubscription(subscription, {
  // subscription options
});
```

We could add a `skip` option for both `useQuery` and `useSubscription` to start them conditionally (#21).

seeden:
@trojanowski  I am fine with useSubscription. I think that many users will not use subscriptionResult. They will use just updateQuery from subscription options

seeden:
Actually updateQuery is related to useQuery. I need to combine useQuery and useSubscription.

seeden:
@trojanowski I was thinking about it and I recommended to use both
1. forward subscribeToMore for queries because it is related to current query.
2. add useSubscription as well. If user want to use just response from subscribe without updating query

src/index.d.ts
```diff
@@ -39,6 +40,13 @@ type QueryHookOptions<TVariables> = Omit<QueryOptions<TVariables>, 'query'> & {
   suspend?: boolean;
 };
 
+type SubscribeToMoreHookOptions = Omit<
+  SubscribeToMoreOptions<TData, TVariables, TSubscriptionData>,
+  'updateQuery'
+> & {
+  updateQuery: UpdateQueryFn<TData, TVariables, TSubscriptionData>;
```
dimitar-nikovski: updateQuery is market as optional in the source, however, without it, subscribeToMore has no side effects, [see code](https://github.com/apollographql/apollo-client/blob/911ca2db631adc38745b8098dd802e0cdfe8dbbb/packages/apollo-client/src/core/ObservableQuery.ts#L367)

src/index.js
```diff
@@ -47,6 +54,17 @@ export function useQuery(query, { suspend = true, ...restOptions } = {}) {
     refetch: variables => observableQuery.current.refetch(variables),
     startPolling: interval => observableQuery.current.startPolling(interval),
     stopPolling: () => observableQuery.current.stopPolling(),
+    subscribeToMore: opts => {
+      const { document, variables } = opts;
+      if (
+        document !== previousSubscription.current ||
+        isEqual(variables, previousSubscribtionOptions.current)
+      ) {
+        unsubscribeToMore.current = observableQuery.current.subscribeToMore(
```
dimitar-nikovski: `subscribeToMore` returns a function to unsubscribe the subscription [see here](https://github.com/apollographql/apollo-client/blob/911ca2db631adc38745b8098dd802e0cdfe8dbbb/packages/apollo-client/src/core/ObservableQuery.ts#L390)

dimitar-nikovski:
Hi @seeden, now it should be much better, as @trojanowski pointed out, there are necessary conditions to set it up properly. I don't expect it to be merged, unless there is genuine desire from more people for it to be part of the project, for me, it mainly helps with the convenience of merging the subscription updates back into the store with `updateQuery`.

- [x] - unsubscribe from the subscription in `useEffect` if it has been set up
- [x] - only resubscribe if the variables or document (gql) have changed, also the recommended way, [source](https://github.com/apollographql/apollo-client/blob/911ca2db631adc38745b8098dd802e0cdfe8dbbb/packages/apollo-client/src/core/ObservableQuery.ts#L357)

Conveniently, when merging the results in `updateQuery`, Apollo's ObservableQuery will schedule an update which will broadcast right into the subscription made in [useQuery](https://github.com/trojanowski/react-apollo-hooks/blob/e1a9b6f77d2f9fb8c34ad89ee0ec5e3720e1b16d/src/index.js#L31), therefore then setting a new result as if it was a change in query result, so that would then become query + subscription merged.

seeden:
Hi @dimitar-nikovski thank you for your update 👍. Actually I am using subscribeToMore only with updateQuery. But it is related to useQuery because I am using old query result in the updateQuery (it is standard in the react-apollo and Query component I guess). I am merging edges most of the time.

dimitar-nikovski:
I ought to add some tests, just in case it will be going further.

src/index.js
dimitar-nikovski: conditional mistake corrected in the commit below

trojanowski:
What do you think about the API suggested in https://github.com/trojanowski/react-apollo-hooks/pull/37#issuecomment-450200481? It would allow to update any previous query.

trojanowski:
Do you think it would solve all your use cases for _subscribeToMore_ or you still need it?

dimitar-nikovski:
@trojanowski it's cleaner indeed, we can use the subscription result from useSusbscription and the updateQuery prop form ObservableQuery returned by useQuery and achieve the exact same result. Feel free to close this down.

PR #37, opened 2018-12-18
add useSubscription

### LINKED ISSUE #22: useSubscription

sijad:
fixes #22 

TODO:

- [x] add tests
- [ ] add document to README

### DIFF
--- a/src/__tests__/useSubscription-test.tsx
+++ b/src/__tests__/useSubscription-test.tsx
@@ -1,0 +1,329 @@
+import { ApolloClient } from 'apollo-client';
+import { Operation } from 'apollo-link';
+import { MockSubscriptionLink } from 'apollo-link-mock';
+import gql from 'graphql-tag';
+import React from 'react';
+import { cleanup, render } from 'react-testing-library';
+
+import { ApolloProvider, useSubscription } from '..';
+import createClient from '../__testutils__/createClient';
+import { SAMPLE_TASKS } from '../__testutils__/data';
+import wait from '../__testutils__/wait';
+
+jest.mock('../internal/actHack');
+
+const TASKS_SUBSCRIPTION = gql`
+  subscription NewTasks {
+    task {
+      id
+      text
+      completed
+    }
+  }
+`;
+
+const FILTERED_TASKS_SUBSCRIPTION = gql`
+  subscription FilteredNewTasks($completed: Boolean!) {
+    task(completed: $completed) {
+      id
+      text
+      completed
+    }
+  }
+`;
+
+const results = SAMPLE_TASKS.map(task => ({
+  result: { data: { task } },
+}));
+
+const complitedResults = SAMPLE_TASKS.filter(task => task.completed).map(
+  task => ({
+    result: { data: { task } },
+  })
+);
+
+const uncomplitedResults = SAMPLE_TASKS.filter(task => !task.completed).map(
+  task => ({
+    result: { data: { task } },
+  })
+);
+
+afterEach(cleanup);
+
+it('should return the subscription data', async () => {
+  const link = new MockSubscriptionLink();
+  const client = createClient({ link });
+
+  let count = 0;
+
+  const Component = () => {
+    const { data, loading, error } = useSubscription(TASKS_SUBSCRIPTION);
+    if (count === 0) {
+      expect(loading).toBe(true);
+      expect(error).toBeUndefined();
+      expect(data).toBeUndefined();
+    } else if (count === 1) {
+      expect(loading).toBe(false);
+      expect(error).toBeUndefined();
+      expect(data).toEqual(results[0].result.data);
+    } else if (count === 2) {
+      expect(loading).toBe(false);
+      expect(error).toBeUndefined();
+      expect(data).toEqual(results[1].result.data);
+    } else if (count === 3) {
+      expect(loading).toBe(false);
+      expect(error).toBeUndefined();
+      expect(data).toEqual(results[2].result.data);
+    }
+    count++;
+    return null;
+  };
+
+  render(
+    <ApolloProvider client={client}>
+      <Component />
+    </ApolloProvider>
+  );
+
+  for (let i = 0; i < 3; i++) {
+    link.simulateResult(results[i]);
+    await wait();
+  }
+
+  expect(count).toBe(4);
+});
+
+it('should return the subscription error', async () => {
+  const link = new MockSubscriptionLink();
+  const client = createClient({ link });
+
+  const subscriptionError = {
+    error: new Error('error occurred'),
+  };
+
+  let count = 0;
+
+  const Component = () => {
+    const { data, loading, error } = useSubscription(TASKS_SUBSCRIPTION);
+    if (count === 0) {
+      expect(loading).toBe(true);
+      expect(error).toBeUndefined();
+      expect(data).toBeUndefined();
+    } else if (count === 1) {
+      expect(loading).toBe(false);
+      expect(error).toEqual(new Error('error occurred'));
+      expect(data).toBeUndefined();
+    }
+    count++;
+    return null;
+  };
+
+  render(
+    <ApolloProvider client={client}>
+      <Component />
+    </ApolloProvider>
+  );
+
+  link.simulateResult(subscriptionError);
+  await wait();
+
+  expect(count).toBe(2);
+});
+
+it('should call provided onSubscriptionData', async () => {
+  const link = new MockSubscriptionLink();
+  const client = createClient({ link });
+
+  let count = 0;
+
+  const Component = () => {
+    useSubscription(TASKS_SUBSCRIPTION, {
+      onSubscriptionData: opts => {
+        const { loading, data, error } = opts.subscriptionData;
+        expect(opts.client).toBeInstanceOf(ApolloClient);
+        expect(data).toEqual(results[count].result.data);
+        expect(loading).toBe(false);
+        expect(error).toBeUndefined();
+        count++;
+      },
+    });
+    return null;
+  };
+
+  render(
+    <ApolloProvider client={client}>
+      <Component />
+    </ApolloProvider>
+  );
+
+  for (let i = 0; i < 3; i++) {
+    link.simulateResult(results[i]);
+    await wait();
+  }
+
+  expect(count).toBe(3);
+});
+
+it('should execute subscription with provided variables', async () => {
+  const variables = { completed: true };
+
+  class MockSubscriptionLinkOverride extends MockSubscriptionLink {
+    request(req: Operation) {
+      expect(req.variables).toEqual(variables);
+      return super.request(req);
+    }
+  }
+
+  const link = new MockSubscriptionLinkOverride();
+
+  const client = createClient({ link });
+
+  let count = 0;
+
+  const Component = () => {
+    const { data, loading, error } = useSubscription(
+      FILTERED_TASKS_SUBSCRIPTION,
+      { variables }
+    );
+    if (count === 0) {
+      expect(loading).toBe(true);
+      expect(error).toBeUndefined();
+      expect(data).toBeUndefined();
+    } else if (count === 1) {
+      expect(loading).toBe(false);
+      expect(error).toBeUndefined();
+      expect(data).toEqual(complitedResults[0].result.data);
+    }
+    count++;
+    return null;
+  };
+
+  render(
+    <ApolloProvider client={client}>
+      <Component />
+    </ApolloProvider>
+  );
+
+  link.simulateResult(complitedResults[0]);
+  await wait();
+
+  expect(count).toBe(2);
+});
+
+it('should not re-subscription if variables have not changed', async () => {
+  const variables = { completed: true };
+
+  class MockSubscriptionLinkOverride extends MockSubscriptionLink {
+    request(req: Operation) {
+      expect(req.variables).toEqual(variables);
+      return super.request(req);
+    }
+  }
+
+  const link = new MockSubscriptionLinkOverride();
+
+  const client = createClient({ link });
+
+  let count = 0;
+
+  const Component = () => {
+    const [, forceRender] = React.useState(0);
+    const { data, loading, error } = useSubscription(
+      FILTERED_TASKS_SUBSCRIPTION,
+      { variables }
+    );
+    if (count === 0) {
+      expect(loading).toBe(true);
+      expect(error).toBeUndefined();
+      expect(data).toBeUndefined();
+    } else if (count === 1) {
+      expect(loading).toBe(false);
+      expect(error).toBeUndefined();
+      expect(data).toEqual(complitedResults[0].result.data);
+      forceRender(c => c + 1);
+    } else if (count === 2) {
+      expect(loading).toBe(false);
+    }
+    count++;
+    return null;
+  };
+
+  render(
+    <ApolloProvider client={client}>
+      <Component />
+    </ApolloProvider>
+  );
+
+  link.simulateResult(complitedResults[0]);
+  await wait();
+
+  expect(count).toBe(3);
+});
+
+it('should re-subscription if variables have changed', async () => {
+  class MockSubscriptionLinkOverride extends MockSubscriptionLink {
+    variables: any;
+    request(req: Operation) {
+      this.variables = req.variables;
+      return super.request(req);
+    }
+
+    simulateResult() {
+      if (this.variables.completed) {
+        return super.simulateResult(complitedResults[0]);
+      } else {
+        return super.simulateResult(uncomplitedResults[0]);
+      }
+    }
+  }
+
+  const link = new MockSubscriptionLinkOverride();
+
+  const client = createClient({ link });
+
+  let count = 0;
+
+  const Component = () => {
+    const [completed, setCompleted] = React.useState(false);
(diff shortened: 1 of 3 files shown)


trojanowski:
@sijad Thanks for this PR. It would be great if you also added some tests.

From your to-do list:

>  support suspend

I'm not sure if it's necessary or how would it work? I think suspense is great for loading required data, but not necessary for notifications use cases.

DAB0mB:
Here's a gist that I created that include TypeScript definitions. Note that instead of using `null` for `data` I've used an empty object. It's much more convenient to work with it this way, since you're always gonna pluck the root field: https://gist.github.com/DAB0mB/6e3d3e2e9470b89c68365c981f9d0a17.

Here's an example usage of an app I'm creating:

```ts
const { chatId, content } = options.variables
const { data: { messageAdded } } = useSubscription<MessageAdded.Subscription, MessageAdded.Variables>(messageAddedSubscription, {
  variables: { chatId }
})

useEffect(() => {
  if (!messageAdded) return

  const { chat }: GetChat.Query = store.readQuery({
    query: getChatQuery,
    variables: { chatId },
  })

  chat.messages.push(messageAdded)

  store.writeQuery({
    query: getChatQuery,
    variables: { chatId },
    data: { chat },
  })

  const { chats }: GetChats.Query = store.readQuery({
    query: getChatsQuery,
  })

  const chat = chats.find(chat => chat.id === chatId)

  if (chat) {
    chat.messages.push(messageAdded)

    store.writeQuery({
      query: getChatsQuery,
      data: { chats },
    })
  }
}, [messageAdded && messageAdded.id])
```

DAB0mB:
What do you guys think of the following API? This way we can skip the `useEffect()` and cut to the chase:

```
useSubscription(SUBSCRIPTION_OR_OPTIONS, ({ data, error, loading }) => {
  // Do something...
})
```

This will be invoked whenever new data is emitted.

trojanowski:
What do you think about this API instead? It would be similar to the `onSubscriptionData` callback from the `<Subscription />` component from react-apollo: <https://github.com/apollographql/react-apollo/blob/99a350a0f504c35dedee8ebe73dd1d04e2a97c23/src/Subscriptions.tsx#L31>:

```javascript
const { data, error, loading } = useSubscription(SUBSCRIPTION, ({ client, subscriptionData }) => {
  // Do something...
})
```

DAB0mB:
Having a callback is the way to go, no doubt. But why do we need the returned `data, error, loading` outside of it? I'm working on a pretty big app and didn't have any use for them.

sijad:
I think callback is better, but I'm not sure about `SUBSCRIPTION_OR_OPTIONS`.

sijad:
if we go with callback, user need to use `useState` to access to `error` and `loading`, anyway here's the [callback implementation](https://gist.github.com/sijad/ebda90e898b244bf86b9bbbf96fb291c#file-usesubscriptionwithcallback-js):

trojanowski:
I'd like to update my previous suggestion to that one:

```typescript
const { data, error, loading } = useSubscription(MY_SUBSCRIPTION, {
  variables: {
    // ...
  },
  onSubscriptionData: ({ client, subscriptionData }) => {
    // optional callback similar to the one from react-apollo:
    // https://www.apollographql.com/docs/react/advanced/subscriptions.html#props
  }
  // ... rest options
});
```

In some cases, you're interested in just the last subscription value sent by the server (e. g. global indicator showing how many new messages you have in an instant messenger app). In that case, it would be enough just to look at the `data` property returned by the hook. In other cases it's not enough - e. g. you'd like to show a notification to the user or modify the Apollo cache - e. g. you'd like to show a new comment on a blog post page for a user visiting it just after it was created - `onSubscriptionData` callback would be helpful then.

sijad:
I just pushed TS version, `onSubscriptionData` also implemented but I'm not sure about passing `client`, `useApolloClient` is easy enough to use.

I'll add some tests soon.

matthewcarlreetz:
Howdy. I'm thinking about using this lib with a project I'm currently working on. Curious what the status of `useSubscription` is as it is a feature I'd like to utilized. Thanks!

sijad:
I'm sorry for my delay, I'll add tests in few days.

src/useSubscription.ts
```diff
@@ -0,0 +1,100 @@
+import {
+  ApolloClient,
+  ApolloError,
+  OperationVariables,
+  SubscriptionOptions,
+} from 'apollo-client';
+import { DocumentNode } from 'graphql';
+import { useEffect, useRef, useState } from 'react';
+import isEqual from 'react-fast-compare';
```
bradleyayers: `react-fast-compare` isn't currently a dependency, can we get this from an existing dependency, or should do you want to add `react-fast-compare` as a dependency?

seeden:
any update guys?

trojanowski:
@trojanowski I'll try to add support for subscriptions next week.

trojanowski:
@seeden  I'll try to add support for subscriptions next week.

bradleyayers:
I've just copied this into my project and have been using it successfully. I've also used it in my own `useQuery2` that exposes the subscription hook:

```ts
export type UseQuery2UpdateQuery<TData, TSubscriptionData> = (
  previousQueryResult: TData,
  options: {
    subscriptionData: SubscriptionHookResult<TSubscriptionData>;
  }
) => TData;

/**
 * Wrapped version of `useQuery` that fixes the return type to include
 * `undefined`, and exposes a `useSubscribeToMore` hook.
 */
export const useQuery2 = <TData, TVariables>(query: DocumentNode, opts?: QueryHookOptions<TVariables>) => {
  const result: QueryHookResult<TData, TVariables> = useQuery(query, opts);

  // Using a ref here allows the `onSubscriptionData` closure to access the latest value.
  const resultRef = useRef(result);
  resultRef.current = result;

  return {
    ...result,
    data: result.data as Partial<TData>,
    useSubscribeToMore: <TSubscriptionData, TSubscriptionVariables>(
      subscriptionQuery: DocumentNode,
      subscriptionOptions: {
        variables: TSubscriptionVariables;
        fetchPolicy?: FetchPolicy;
        updateQuery: UseQuery2UpdateQuery<TData, TSubscriptionData>;
        // `updateQuery` is implicitly memoised, and mandating this param forces
        // the consumer to think about that.
        updateQueryInputs: ReadonlyArray<unknown>;
      }
    ) => {
      const { updateQuery, updateQueryInputs, ...passThroughOptions } = subscriptionOptions;
      useSubscription<TSubscriptionData, TSubscriptionVariables>(subscriptionQuery, {
        ...passThroughOptions,
        onSubscriptionData: useCallback<OnSubscriptionData<TSubscriptionData>>(({ subscriptionData }) => {
          resultRef.current.updateQuery(previousQueryResult =>
            updateQuery(previousQueryResult, {
              subscriptionData
            })
          );
        }, updateQueryInputs)
      });
    }
  };
};
```

Perhaps that'll be useful to someone.

seeden:
Thanks, @bradleyayers I am going to test it as well. You made my day.

Pruxis:
> @seeden I'll try to add support for subscriptions next week.

A release date in mind?

seeden:
@trojanowski please add it :)

bradleyayers:
After using this in practice, one thing that's missing here is the ability to conditionally call `useSubscription`. I've needed this for a few reasons:

- When the query returns an error and I'm not using suspense
- When the subscription is only needed when the query result includes optional fields. A specific example is that my query may return an `image` field (it's optional), and if it does, I need to subscribe to changes to that image.

In code, this looks like:

```ts
const { data, error } = useQuery(…);

if (data && !error) {
  useSubscription(…, { variables: { id: data.image.id });
}
```

Perhaps it should look more like:

```ts
const { data, error } = useQuery(…);

useSubscription(data && !error ? { query: …, variables: { id: data.image.id } : null);
```

Having the ability to pass `null` into `useSubscription` would allow it to be called unconditionally.

Has anyone else faced this or seen a similar problem solved in other libraries?

bradleyayers:
What's the use-case for `shouldResubscribe` that can't be handled by just looking at the query and variables?

src/useSubscription.ts
```diff
@@ -0,0 +1,100 @@
...
+    inputs = [];
+  } else if (shouldResubscribe === undefined) {
+    inputs = [
+      query,
+      isEqual(prevOptions.current, options) ? prevOptions.current : options,
+    ];
+  }
+
+  useEffect(() => {
+    if (options) {
```
bradleyayers: Why is this conditional? How does this handle the case where `useSubscription` might be called multiple times where `options` is only sometimes provided?

sijad: If shouldResubscribe is undefinded input would be equal to an array contains query and options which means useEffect only resubsribes only if one of those values has been changed, if shouldResubscribe was equal to false then input would be an empty array which means useEffect never resubscribes, other wise input would be undefined and useEffect will resubscribes with every call.

bradleyayers: Something else worth pointing out here is that `useEffects` requires the deps (second) parameter to always be the same size array. It will throw an error if you pass an empty array one time, and an array of length 2 another. As such I think line 61 should be `[null, null]` or something similar.

sijad: it that case I think there would be better solution here, currently it's ugly as it is.

trojanowski:
@seeden The main issue which prevents me from merging this PR is lack of tests.
@sijad I also have some comments/questions:

1. What is the use case for the `shouldResubscribe` option? Isn't it enough to resubscribe if mutation or options are changed?
2. Instead of using _react-fast-compare_ you could use `React.memo` similar to how it's done here: https://github.com/trojanowski/react-apollo-hooks/blob/39f59eba268a9b1725bae9ab32efc1af740d506d/src/useQuery.ts#L89

Unfortunately, I wasn't able to work on it last week, but I'll try in the next few days - to add tests and prepare an alternative PR (or to review and merge one if it will be available with tests :) ).

@bradleyayers For skipping queries we could use the `skip` option, similar to the one in `useQuery` (undocumented, unfortunately), e. g. such invocation would prevent a subscription from starting:

```javascript
useSubscription(MY_SUBSCRIPTION, { skip: true });
```

seeden:
@trojanowski thanks I will wait for your version 👍

bradleyayers:
> For skipping queries we could use the `skip` option, similar to the one in `useQuery` (undocumented, unfortunately), e. g. such invocation would prevent a subscription from starting:

That sounds okay to me, and being consistent with `useQuery` is great. I always specific the generic types when calling the function, so to work with that in practice it'd probably end up looking like:

```ts
useSubscription<MyQuery, MyQueryVariables>(
  MY_QUERY,
  data != null
    ? { variables: { id: data.foo.id } }
    : { skip: true }
);
```

or 

```ts
if (data != null) {
  useSubscription<MyQuery, MyQueryVariables>(MY_QUERY, { variables: { id: data.foo.id } });
} else {
  useSubscription<MyQuery, MyQueryVariables>(MY_QUERY, { skip: true });
}
```

but this would violate lint rules that detect hooks used in conditional branches, so I'd go with the first option.

sijad:
1. I've removed `shouldResubscribe` for now, I don't need it either.
2. now it uses `objToKey` instead of `isEqual`.

p.s. please consider that `objToKey` is very slow (https://jsperf.com/objtokey-vs-react-fast-compare/1) and consuming way more memory

src/useSubscription.ts
```diff
@@ -0,0 +1,86 @@
...
+export function useSubscription<TData = any, TVariables = OperationVariables>(
+  query: DocumentNode,
+  {
+    variables,
+    onSubscriptionData,
+    ...options
+  }: SubscriptionHookOptions<TData, TVariables> = {}
+): SubscriptionHookResult<TData> {
+  const client = useApolloClient();
+  const onSubscriptionDataRef = useRef<null | OnSubscriptionData<TData>>(null);
```
kamalbennani: Don't we need to expose unsubscribe method?

KevinKelbie:
Whats holding back merging this Pull Request? Is it functional?

seeden:
When you add it I can completely remove react-apollo from my code.

FredyC:
@sijad Can you please fix https://github.com/trojanowski/react-apollo-hooks/pull/37#pullrequestreview-203494877?

@trojanowski Is there something else to be done before merging it?

sijad:
@fredyc it's fixed already

src/__tests__/useSubscription-test.tsx
```diff
@@ -0,0 +1,349 @@
+import { ApolloClient } from 'apollo-client';
```
trojanowski: It would be nice to merge master into this PR. After that the following changes will be required:

1. `flushEffects` won't be required any more and it will be possible to remove all invocations of it.
2. `flushEffectsAndWait` should be replaced with `wait` from `../__testutils__/wait`.
3. You could add `jest.mock('../internal/actHack');` just below the imports to apply the _act hack_ (please look at the review of _useSubscription-test.tsx_ for more details).

After applying those changes it could look like that file: https://gist.github.com/trojanowski/9bb29c77537371919fd861cd47cb5015#file-usesubscription-test-tsx

src/useSubscription.ts
trojanowski: I'd change it to:

``` 
const onSubscriptionDataRef = useRef<
    OnSubscriptionData<TData> | null | undefined
  >(null);
```

to allow `onSubscriptionData` be both `null` and `undefined`.

src/useSubscription.ts
```diff
@@ -0,0 +1,86 @@
...
+  const [result, setResult] = useState<SubscriptionHookResult<TData>>({
+    loading: true,
+  });
+
+  if (onSubscriptionData) {
```
trojanowski: I don't think this `if` is correct. In my opinion `onSubscriptionDataRef.current = onSubscriptionData;` should always be invoked. Otherwise, it's possible that the old `onSubscriptionData` callback will be invoked if it is replaced with `null`. E. g. please consider this snippet:

```typescript
const [aValue, setValue] = useState(1);
useSubscription(MY_SUBSCRIPTION, {
  onSubscriptionData: (aValue === 1) ? someCallback : null,
});
```

src/useSubscription.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  useEffect(
+    () => {
+      const subscription = client
+        .subscribe({
+          ...options,
+          query,
+          variables,
+        })
+        .subscribe({
```
trojanowski: `error` before `next` looks a little unusual for me. I guess it's that way because of a tslint rule. So I'd suggest using a different form of `subscribe` here:

```
       .subscribe(
          nextResult => {
            // ...
          },
          error => {
            // ...
          }
        );
```

It's also the same form used in `useQuery` and it should allow us to save a byte or two in the final bundle 😉.

src/useSubscription.ts
```diff
@@ -0,0 +1,86 @@
...
+          error: error => {
+            setResult({ loading: false, data: result.data, error });
+          },
+          next: nextResult => {
+            const newResult = {
+              data: nextResult.data,
+              error: undefined,
+              loading: false,
+            };
+            setResult(newResult);
```
trojanowski: Using `setResult` here causes React warnings. Until https://github.com/facebook/react/issues/14769 is fixed I recommend a workaround similar to the one in `useQuery`: https://github.com/trojanowski/react-apollo-hooks/blob/6c9c90fa5fcb349b3e3e308bfb93901e59af6a3a/src/useQuery.ts#L157

In this file it could be applied that way:

1. I'd rename the function for setting state returned  by `useState` to e. g. `setResultBase`.
2. I'd define `setResult` like this:

```
  function setResult(newResult: SubscriptionHookResult<TData>) {
    actHack(() => {
      setResultBase(newResult);
    });
  }
```

`actHack` is available with `import actHack from './internal/actHack';`.

After https://github.com/facebook/react/pull/14853 we should be able to get rid of those hacks.

src/useSubscription.ts
```diff
@@ -0,0 +1,86 @@
...
+              });
+            }
+          },
+        });
+      return () => {
+        setResult({ loading: true });
+        subscription.unsubscribe();
+      };
+    },
+    [query, variables && objToKey(variables)]
```
trojanowski: I'd add `options && objToKey(options)` here just to be sure.

trojanowski:
Sorry, it took so long. I didn't have too much time in the last few days. I added my comments. Feel free to discuss them.

src/useSubscription.ts
FredyC: Why is that a good thing? It only makes the use more annoying as you need to check for both separately.

src/useSubscription.ts
FredyC: Feels like edge case why would someone want conditionally set the callback while it's easier to check the condition inside that callback. Doesn't mean it shouldn't be fixed though :)

sijad:
@trojanowski thanks for review, all of suggestions implemented. I just removed `variables && objToKey(variables)` due the fact `variables` is a member of `options`. am I correct?

trojanowski:
Thank you for your work @sijad. In a moment I'll add some documentation and publish a new release.

> I just removed `variables && objToKey(variables)` due the fact `variables` is a member of `options`. am I correct?

Yep.

trojanowski:
I just published a version 0.4.1 with the `useSubscription` hook.

seeden:
Thanks @trojanowski for great job

PR #38, opened 2018-12-19
feat(useQuery): Implement `skip` prop.

### LINKED ISSUE #21: Conditionally firing a query

umidbekkarimov:
Closes #21

### DIFF
--- a/src/__tests__/useQuery-test.js
+++ b/src/__tests__/useQuery-test.js
@@ -112,6 +112,7 @@ function TaskList({ tasks }) {
 
 function TasksLoader({ query, ...restOptions }) {
   const { data, error } = useQuery(query, restOptions);
+
   if (error) {
     throw error;
   }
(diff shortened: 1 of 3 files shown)


trojanowski:
@umidbekkarimov Is there a reason why you closed it? I like this PR and would like it merged.

umidbekkarimov:
@trojanowski it introduced  new bug:

```js
const { accessToken } = useAuthToken();

const { data, errors, loading, refetch } = useQuery(MeQuery, {
  skip: !accessToken,

  suspend: false,
  errorPolicy: "all",
  fetchPolicy: "cache-first"
});

useEffect(
  () => {
    if (accessToken) {
      refetch(); // throws because observableQuery.current not defined yet.
    }
  },
  [accessToken]
);
```

I rewrote whole `useQuery` with `useMemo` usage (https://github.com/umidbekkarimov/react-apollo-hooks/blob/memo/src/useQuery.ts)

PR #39, opened 2018-12-19
refactor: Rewrite in TypeScript.

umidbekkarimov:


trojanowski:
Thanks @umidbekkarimov for your great work here. I'll do a full review after Christmas, but right now I have 2 suggestions:

1. I'd like to have the git history and changelog (generated by [standard-version](https://github.com/conventional-changelog/standard-version)) easy to follow so I'd like to merge it using the "Squash and merge" option. It would be great if this PR just renames the current files and adds TS types there (so _index.js_ -> _index.ts_ etc.), and splitting hooks to separate files is done in a follow-up PR.
2. I'd use [TSLint](https://palantir.github.io/tslint/) instead of ESLint for linting the TypeScript code. It seems more popular for that case and it's also already used in Apollo repositories. I'm using this configuration in my projects (maybe it needs some customization):

```json
{
  "defaultSeverity": "error",
  "extends": ["tslint:latest", "tslint-react", "tslint-config-prettier"],
  "jsRules": {},
  "rules": {
    "interface-name": false,
    "interface-over-type-literal": false,
    "jsx-boolean-value": false,
    "jsx-no-lambda": false,
    "max-classes-per-file": false,
    "member-access": false,
    "no-implicit-dependencies": true,
    "no-submodule-imports": false,
    "no-this-assignment": [true, { "allow-destructuring": true }],
    "no-unused-expression": [true, "allow-fast-null-checks"],
    "only-arrow-functions": false,
    "ordered-imports": [
      true,
      {
        "import-sources-order": "lowercase-last",
        "named-imports-order": "lowercase-last"
      }
    ],
    "prefer-template": true
  },
  "rulesDirectory": []
}
```

Please let me know if you don't have time for that - I'd do it myself in the next week in that case.

umidbekkarimov:
> I'd like to have the git history and changelog (generated by standard-version) easy to follow so I'd like to merge it using the "Squash and merge" option. It would be great if this PR just renames the current files and adds TS types there (so index.js -> index.ts etc.), and splitting hooks to separate files is done in a follow-up PR.

Done, I tried save git history as much as possible in force pushed commits, but git still wants to remove and add `index.tsx` and `objToKey.ts` files.

> I'd use TSLint instead of ESLint for linting the TypeScript code. It seems more popular for that case and it's also already used in Apollo repositories. I'm using this configuration in my projects (maybe it needs some customization):

Done.

> I'll do a full review after Christmas

Happy Holidays!

src/__testutils__/noop.ts
```diff
@@ -0,0 +1,3 @@
+export function noop() {
```
trojanowski: `export default function noop()` ? There is only one export and it has the same name os the file.

src/index.tsx
```diff
@@ -1,37 +1,103 @@
...
 import objToKey from './objToKey';
 import {
   getCachedObservableQuery,
   invalidateCachedObservableQuery,
 } from './queryCache';
 
-const ApolloContext = React.createContext();
+const ApolloContext = React.createContext<null | ApolloClient<any>>(null);
+
+export interface ApolloProviderProps {
```
trojanowski: Shouldn't it be parametrized like in the old definitions: <https://github.com/trojanowski/react-apollo-hooks/blob/e1a9b6f77d2f9fb8c34ad89ee0ec5e3720e1b16d/src/index.d.ts#L25>?

tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "include": ["src/**/*.ts", "src/**/*.tsx"],
+  "compilerOptions": {
+    "jsx": "react",
+    "moduleResolution": "node",
+    "allowSyntheticDefaultImports": true,
```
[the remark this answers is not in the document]
umidbekkarimov: Now will break usage of `lodash/isPlainObject` and `react-fast-compare`. So we will have to use `require` syntax for these cases.

src/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+import React, {
+  ReactElement,
+  ReactNode,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+import isEqual from 'react-fast-compare';
+import { Omit, assertApolloClient } from './Utils';
```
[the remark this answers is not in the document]
umidbekkarimov: Maybe this is the solution https://github.com/renke/import-sort? I has similar api as prettier: `--write` and `--list-different`.

src/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+      return () => {
+        subscription.unsubscribe();
+      };
+    },
+    [query, objToKey(restOptions)]
+  );
+
+  ensureSupportedFetchPolicy(suspend, restOptions.fetchPolicy);
+
+  // TODO: https://github.com/apollographql/react-apollo/blob/5cb63b3625ce5e4a3d3e4ba132eaec2a38ef5d90/src/Query.tsx#L54-L59
```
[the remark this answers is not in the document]
umidbekkarimov: #42 wasn't there yet when I add this comment.

src/__testutils__/noop.ts
umidbekkarimov: I used to follow this: https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html and using named exports is just an old habit for me.

tsconfig.json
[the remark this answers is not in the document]
trojanowski: Good catch. It could break some webpack config and possible future node.js mjs integration, so let's better leave it as `true`.

umidbekkarimov:
@trojanowski Thanks, It's good to have you back! I've resolved your comments.

I hate to bring it up, code style changes (e.g. sorting of imports) should be done by tools, not humans. Prettier spoiled me.

src/index.tsx
```diff
@@ -51,7 +51,11 @@ export function ApolloProvider<TCacheShape = any>({
 export function useApolloClient<TCache = object>(): null | ApolloClient<
   TCache
 > {
-  return useContext(ApolloContext);
+  const client = useContext(ApolloContext);
+
+  assertApolloClient(client);
```
trojanowski: This change is potentially breaking, but I think it's a good one. I'd even go one step further - inline `assertApolloClient` here and remove `null` as a possible result of  `useApolloClient` (because after that there would always be a correct instance of the client returned or an error is thrown). It would also allow us to replace all occurrences of `client!` with just `client`.

trojanowski:
Thank you @umidbekkarimov 

I added one comment about the new changes.

> I hate to bring it up, code style changes (e.g. sorting of imports) should be done by tools, not humans. Prettier spoiled me.

I agree. Thanks for the tip about [import-sort](https://github.com/renke/import-sort) - I'll try to add it.

umidbekkarimov:
> I added one comment about the new changes.

@trojanowski I can't find any, maybe you forgot to submit it?

src/index.tsx
trojanowski: I'd just inline the `assertApolloClient` function here and remove it from _utils.ts_. It's used only here and that way we wouldn't have to return `client!`, just `client`.

```suggestion
   if (!client) {
    // https://github.com/apollographql/react-apollo/blob/5cb63b3625ce5e4a3d3e4ba132eaec2a38ef5d90/src/component-utils.tsx#L19-L22
    throw new Error(
      'Could not find "client" in the context or passed in as a prop. ' +
        'Wrap the root component in an <ApolloProvider>, or pass an ' +
        'ApolloClient instance in via props.'
    );
  }
```

umidbekkarimov: Agree, I also removed `utils.ts`

.gitignore
```diff
@@ -1,5 +1,6 @@
 /es
 /node_modules
 /lib
+/typings
```
trojanowski: I think that change is no longer necessary - we could use the _typings_ folder to store definitions for external modules (e. g. I think it's a good place to store that file: <https://github.com/trojanowski/react-apollo-hooks/pull/46/files#diff-352f02e2ca9a9158d200b6b51f0e1c4c>). The same for _.prettierignore_
```suggestion
```

trojanowski:
I added one last-minute comment. I think it's the last one :)

umidbekkarimov:
No problem, it's never easy to review 20+ file PRs. Fixed.

umidbekkarimov:
Forgot to push 🤦‍♂️

Issue #36, opened 2018-12-17
useQuery causes a component to render multiple times


asfwebmaster:
Is it normal with this new hooks api to re-render a component multiple times? I just saw that `useQuery` causes a component to re-render 3 times. Here is an example: [https://codesandbox.io/s/j7kw103525](https://codesandbox.io/s/j7kw103525)

It's the same, example from the readme file, I have just added a console log to /views/Details.js

```
const Detail = ({
  match: {
    params: { breed, id }
  }
}) => {
  const { data, error } = useQuery(GET_DOG, {
    variables: { breed }
  });
  console.log(2);
  return (
    <View style={styles.container}>
      <Header text={breed} />
      {error ? (
        <Error />
      ) : (
        <DogList...
```
And the output is:
2
2
2

I have tried on my own project and got the same result.

fbartho:
Render with cache-data, render while loading, render with data-result = 3 times. -- That makes sense to me! How many times would you expect it to render? @asfwebmaster

asfwebmaster:
Hm, I have just updated with: `console.log(data, error, loading);`
and the result is: 

```
Object {dog: Object}
dog: Object
images: Array[169]
__typename: "Dog"
undefined
false
---------------------
Object {dog: Object}
dog: Object
images: Array[169]
__typename: "Dog"
undefined
false
---------------------
Object {dog: Object}
dog: Object
images: Array[169]
__typename: "Dog"
undefined
false
```
So the `loading` part is `false` and all other props are same for all renders, this seems a bit strange.

adjourn:
@trojanowski 
I tried to figure it out but I got stuck, is several renders due to `react-apollo-hooks` or `apollo-client`?

2 renders seems a bit overkill e.g when only querying `@client` data (which is already in cache) and `loading` is `false` both times. I double checked that component renders once without `useQuery`.

Component renders once when `skip: true` option is used.

adjourn:
It seems that `react-apollo` had similar issue https://github.com/apollographql/react-apollo/pull/1849.

Are you planning to address this issue in `react-apollo-hooks` soon or ever? It has a pretty nasty effect if you e.g query `locale` at top level and other values in different parts, re-renders cascade all over the app.

mbrowne:
It looks like this issue has something to do with suspense. When doing:

```js
  <Suspense fallback={<div>Loading...</div>}>
    <MyComponent />
  </Suspense>
```

[...]

Issue #33, opened 2018-12-14
return memoized value


sijad:
currently returned value both of `useMutation` and `useQuery` calculated every time they called, I'm not really familiar with apollo logics, but I guess this can memoized?, for instance `useMutation` should be similar like this:

```js
export function useMutation(mutation, baseOptions) {
  const previousOptions = useRef();
  const client = useApolloClient();
  const result = isEqual(previousOptions.current, baseOptions);
  return useMemo(
    () => {
      previousOptions.current = baseOptions;
      return localOptions => client.mutate({
        mutation,
        ...baseOptions,
        ...localOptions,
      });
    },
    [
      mutation,
      result ? previousOptions.current : baseOptions,
    ],
  );
}
```

trojanowski:
@sijad Have you tried to profile it if the version with memoization is really faster than the current one?

sijad:
no, I did not, but I guess with memorized version you can actually prevent re-rendering e.g:

```js
const mutate = useMutation(MUTATION)
const memoizedMutation = useCallback(() => {
  mutate({
    variables,
  });
}, [mutate]);

return (<button onClick={memoizedMutation}>click on me</button>);
```

with current implimention `mutate` always have a new value therefore it might re-render more...

umidbekkarimov:
We also can `useRef` to imitate `this` in class components:

```ts
export function useMutation<TData, TVariables = OperationVariables>(
  mutation: DocumentNode,
  baseOptions?: MutationHookOptions<TData, TVariables>
): MutationFn<TData, TVariables> {
  const client = useApolloClient();

  // Use refs to access variable values inside callback.
  const mutationRef = useRef(mutation);
  const baseOptionsRef = useRef(baseOptions);

  // Update references on each render.
  mutationRef.current = mutation;
  baseOptionsRef.current = baseOptions;

  // Memoize callback.
  const mutationFnRef = useRef<MutationFn<TData, TVariables>>(options =>
    client.mutate({
      mutation: mutationRef.current,
      ...baseOptionsRef.current,
      ...options,
    })
  );

  return mutationFnRef.current;
}
```

rovansteen:
A downside of not memoizing the value of `useMutation` or `useQuery` is that they break the memoization of hooks that are built on top of them if you follow the rules (and eslint plugin) for hooks strictly.

sijad:
[...]

Issue #29, opened 2018-12-04
Network error is not reported in errors and loading stays true


shlomokraus:
When useing `useQuery` (with suspense false), I get a network error, but the hook state stays with "loading: true" and errors undefined. networkState is still 1.

umidbekkarimov:
You can use `errorPolicy=all` as a workaround.
I guess the problem is that subscription error handler not passed here  https://github.com/trojanowski/react-apollo-hooks/blob/master/src/index.js#L31

mateusznowosielski:
I have exactly the same issue as @shlomokraus described. Changing `errorPolicy='all'` doesn't make any difference - I tried to set that for individual query and globally in ApolloClient level. With `<Suspense>` it works as expected. I'm using version 0.2.1.

mateusznowosielski:
Thanks @umidbekkarimov this is very helpful. Error gets properly handled, I also checked the `skip` option and I like the new typescript aspect. Hope it gets merged.

trojanowski:
I think it will be merged soon - I'm going to review it tomorrow.

cescox:
Hello, any news on this issue?

PR #42, opened 2018-12-21
feat(useQuery): Implement `skip`.

### LINKED ISSUE #21: Conditionally firing a query

umidbekkarimov:
Lot of work done here in order to make `skip` work as expected.

1. `useRef` replaces with `useMemo`.
2. Result is taken from `observableQuery.currentResult` and memoized, so it's always synced and do not requires extra rerenders (also removed `react-fast-compare`)
3. Tests are refactored and `jest-react-profiler` used to track render number.

Also this PR depends on changes from #39.
____
Potentially related to #36 -because of removal of extra `setResults` calls.
Simple memoization (without deep traversal checks) related to #33 (it's also easy to memoize handlers, if it's worth it ofc.)
Potentially related to #29 (added error listener to query observable subscription)
Closes #21.
____
Sorry @trojanowski  for GIANT PRs, but i've desided to use hooks in new project and couldn't wait for approvals.
All this code is published under `@dc0de/react-apollo-hooks` and used in growing app.


umidbekkarimov:
@trojanowski ready for review

src/useQuery.ts
```diff
@@ -115,11 +117,15 @@ export function useQuery<TData = any, TVariables = OperationVariables>(
         partial: result.partial,
       };
     },
-    [responseId, observableQuery]
+    [skip, responseId, observableQuery]
```
FredyC: Wondering, if this is really necessary. If a query suddenly becomes skipped, the current result won't change. Why would it?

src/useQuery.ts
```diff
@@ -145,6 +151,16 @@ export function useQuery<TData = any, TVariables = OperationVariables>(
     updateQuery: observableQuery.updateQuery.bind(observableQuery),
   };
 
+  if (skip) {
+    // Taken from https://github.com/apollographql/react-apollo/blob/5cb63b3625ce5e4a3d3e4ba132eaec2a38ef5d90/src/Query.tsx#L376-L381
+    return {
+      ...helpers,
```
trojanowski: I don't think generic helpers should be returned here (or even created before) - they could be no-functional. I'd create a separate version of helpers which always throw. Please look at https://gist.github.com/trojanowski/f7e4e75659d2de1166274ad8f0d06bb4

src/useQuery.ts
trojanowski: Good question.

umidbekkarimov: This prevents infinite loads in suspense mode when skip goes from `false` to `true`, checkout [this test](https://github.com/trojanowski/react-apollo-hooks/pull/42/files#diff-2db9c4351ea93b397a084c99511cc106R654).

<details><summary>with skip in useMemo</summary>

```
  console.log src/useQuery.ts:128
    { skip: true,
      responseId: 0,
      currentResult:
       { data: {},
         error: undefined,
         errors: undefined,
         loading: true,
         partial: true } }

  console.log src/useQuery.ts:128
    { skip: false,
      responseId: 0,
      currentResult:
       { data: {},
         error: undefined,
         errors: undefined,
         loading: true,
         partial: true } }

  console.log src/useQuery.ts:128
    { skip: false,
      responseId: 0,
      currentResult:
       { data: { tasks: [Array] },
         error: undefined,
         errors: undefined,
         loading: false,
         partial: false } }
```

</details>


<details><summary>without skip in useMemo</summary>

```
   console.log src/useQuery.ts:128
    { skip: true,
      responseId: 0,
      currentResult:
       { data: {},
         error: undefined,
         errors: undefined,
         loading: true,
         partial: true } }

  console.log src/useQuery.ts:128
    { skip: false,
      responseId: 0,
      currentResult:
       { data: {},
         error: undefined,
         errors: undefined,
         loading: true,
         partial: true } }

  console.log src/useQuery.ts:128
    { skip: false,
      responseId: 0,
      currentResult:
       { data: {},
         error: undefined,
         errors: undefined,
         loading: true,
         partial: true } }

  console.log src/useQuery.ts:128
    { skip: false,
      responseId: 0,
      currentResult:
       { data: {},
         error: undefined,
         errors: undefined,
         loading: true,
         partial: true } }
```

</details>

FredyC: Interesting. I kinda thought that with `useMemo` (and similar) it should be about what variables are actually used within the body of it. Looks like this rule of thumb does not always work 😕

src/useQuery.ts
FredyC: How does that work in Apollo? Are they stubbing those helpers too? Also let's please use DEV-only invariant, throwing an error like that can get ugly.

umidbekkarimov: Done, I add test cases.

umidbekkarimov: Oops, it's already merged 😅

trojanowski: > Done, I add test cases.

@umidbekkarimov Could you please create a PR with those changes? Maybe it's worth a further discussion. What's your opinion about it?

umidbekkarimov: @trojanowski I think it's better to keep things like in react-apollo. We just will expose api of observable query, but developers are responsible for their usage. E.g (https://github.com/trojanowski/react-apollo-hooks/pull/38#issuecomment-449369084):

```js
const { accessToken } = useAuthToken();
const { data, errors, loading, refetch } = useQuery(MeQuery, {
  skip: !accessToken
});

useEffect(
  () => {
    if (accessToken) { // Otherwise it will produce http error.
      refetch(); 
    }
  },
  [accessToken]
);
```

FredyC:
@trojanowski Are you planning to release this soon? Would be super helpful :)

trojanowski:
@FredyC published as a part of release 0.3.0.

PR #44, opened 2018-12-23
feat: Implement SSR.

### LINKED ISSUE #8: Suspense and server side rendering
Hello,

I was curious if you knew what the behavior would be like if this package was used with server side rendering.

My understanding is that Suspense is not currently supported when server side rendering. I think it shows loading state for Suspense components. I'm not sure how the interaction will play out when used in combination with the data collector from Apollo Server.

Do you happen to have any insight on this? Thank you.

umidbekkarimov:
Depends on #39 and #42. Tests copied from react-apollo.

Closes #8.


seeden:
Hi guys. please merge it :) I really need this feature

umidbekkarimov:
@trojanowski ready for review

src/useQuery.ts
```diff
@@ -63,12 +66,22 @@ export function useQuery<TData = any, TVariables = OperationVariables>(
...
+    fetchPolicy: actualCachePolicy,
     errorPolicy,
     fetchResults,
   }: QueryHookOptions<TVariables> = {}
 ): QueryHookResult<TData, TVariables> {
   const client = useApolloClient();
+  const ssrManager = useContext(SSRContext);
+
+  // Modify fetch policy for SSR mode.
+  const fetchPolicy =
```
umidbekkarimov: Also a good place for #15

src/useQuery.ts
```diff
@@ -161,10 +175,17 @@ export function useQuery<TData = any, TVariables = OperationVariables>(
     };
   }
 
-  if (suspend && currentResult.partial) {
-    // throw a promise - use the react suspense to wait until the data is
-    // available
-    throw observableQuery.result();
+  if (currentResult.partial) {
+    if (suspend) {
```
trojanowski: I'd change the order here - SSR should work also for queries using the default suspense mode (at least until suspense-ready official React server renderer is available). So rather:

```typescript
  if (currentResult.partial) {
    // Register request only when `ssr: true`.
    if (ssrInUse) {
      ssrManager.register(observableQuery.result());
    }

    else if (suspend) {
      // throw a promise - use the react suspense to wait until the data is
      // available
      throw observableQuery.result();
    }
  }
```

Also tests should check both suspense and non-suspense mode.

umidbekkarimov: Suspense does not work in react-dom

src/useQuery.ts
```diff
@@ -63,12 +66,22 @@ export function useQuery<TData = any, TVariables = OperationVariables>(
...
     errorPolicy,
     fetchResults,
   }: QueryHookOptions<TVariables> = {}
 ): QueryHookResult<TData, TVariables> {
   const client = useApolloClient();
+  const ssrManager = useContext(SSRContext);
+
+  // Modify fetch policy for SSR mode.
+  const fetchPolicy =
+    ssrManager &&
```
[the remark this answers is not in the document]
umidbekkarimov: I did it that way first, but then i changes it because of skip part (can’t add link from mobile 😕) where this cariable will be useless and these two variables will be used again. Also ssrManager.register has to be used with ! because it’s nullable and typescript isn’t smart enough to assert type.

src/useQuery.ts
trojanowski: > Suspense does not work in react-dom

I'm not sure I understand. I changed the code to the one in the snipped above and removed `suspend: false` from `useQuery` invocations in _getMarkupFromTree-test.tsx_ and tests passed. Or do you mean that the `<Suspense />` component is not supported by the current version of react-dom-server renderer? We could workaround it by creating our custom helper component and recommending to use it instead of `<Suspense />` from React:

```typescript
export const SuspenseSsr = (props: SuspenseProps) => {
  const ssrManager = useContext(SSRContext);
  if (ssrManager) {
    return <>{props.children}</>;
  }

  return (<Suspense {...props} />)
}
```

umidbekkarimov: i think it’s better to force developers pass suspend: false if they want to use ssr, otherwise they will have different behavior on ssr and client side.
suspense assures user that data is not null on render, and if ignore suspense on ssr - this can create randomnull pointer errors. Current approach will escalate suspended promises to react-dom and it will give good error measage.

umidbekkarimov: oops, i’m writing from phone and didn’t saw your last message.

but my previous message also covers your comment.

from my experience when you’re writing ssr app - you have to write it ssr first otherwise it would be hacky and buggy. it’s better to wait for officials suspense support from react. if someone REALLY wants to use suspense for components that also rendered on server and deal with all the problem it produeses - they can write their own suspense component

umidbekkarimov: girhub mobile is killing me 🤦‍♂️, it adds teo comments at once and freezes when i want to remove one

trojanowski: > wouldn’t it better to name it “unstable_SuspenseSSR” then 🤔.

👍 

> i like the idea, i’ll try to implement it tonight

Thank you :)

umidbekkarimov: @trojanowski I could't wait 🙂. So I start catching thrown promises in `getMarkupFromTree` and add tests for `suspend: true`.

seeden: I can not wait too 👍 thanks guys

src/__tests__/getMarkupFromTree-test.tsx
```diff
@@ -0,0 +1,506 @@
...
+  return (
+    <TestErrorBoundary>
+      <ApolloProvider client={client}>
+        <UserDetails {...props} />
+      </ApolloProvider>
+    </TestErrorBoundary>
+  );
+}
+
+it('should run through all of the queries that want SSR with suspense', async () => {
```
trojanowski: The tests are actually the same for suspene / non-suspense versions. We could make it more DRY with something like that:

```typescript
[true, false].forEach(suspend => {
  const prefix = suspend ? 'with' : 'without';

  describe(`${prefix} suspense`, () => {
    it('should run through all of the queries that want SSR with suspense', async () => {
      const client = createMockClient();

      await expect(
        getMarkupFromTree({
          renderFunction: renderToString,
          tree: <UserDetailsWrapper client={client} suspend={suspend} />,
        })
      ).resolves.toBe('James');
    });

    // the same for other tests
  });
});
```

Inline snapshots wouldn't work in that case but it shouldnt't be a problem - we were checking simple strings so we could easily replace `toMatchInlineSnapshot` with `toBe`.

src/__tests__/SuspenseSSR-test.tsx
```diff
@@ -0,0 +1,131 @@
...
+  await flushEffectsAndWait();
+
+  expect(container).toMatchInlineSnapshot(`
+<div>
+  James
+</div>
+`);
+});
+
+it('not throws in react-dom/server with suspense', async () => {
```
trojanowski: We could make these tests more DRY - similar to https://github.com/trojanowski/react-apollo-hooks/pull/44/files#diff-aea7562d1833b3cd594caab10263a452R94

src/__tests__/getMarkupFromTree-test.tsx
```diff
@@ -0,0 +1,506 @@
...
+  );
+}
+
+interface UserWrapperProps extends QueryHookOptions<{}> {
+  readonly client: ApolloClient<object>;
+}
+
+function UserDetailsWrapper({ client, ...props }: UserWrapperProps) {
+  return (
+    <TestErrorBoundary>
```
trojanowski: What's the purpose of using `<TestErrorBoundary>` here? Tests are passing also without it.

src/useQuery.ts
trojanowski: @umidbekkarimov Looks great, thanks :) I added a few comments about testing.

src/__tests__/getMarkupFromTree-test.tsx
umidbekkarimov: I used [describe.each](https://jestjs.io/docs/en/api#describeeachtable-name-fn-timeout) instead, and all tests MUST have similar output so `toMatchInlineSnapshot` works just fine.

src/__tests__/getMarkupFromTree-test.tsx
umidbekkarimov: I wanted to test this block https://github.com/trojanowski/react-apollo-hooks/blob/2425cfa86cf0b92c327927fdfb36ac7eb1a32559/src/getMarkupFromTree.tsx#L25-L31

But I guess I found the better way and forgot to remove 🤔

src/__tests__/SuspenseSSR-test.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+describe.each([[true], [false]])('SuspenseSSR with "suspend: %s"', suspend => {
+  it('not throws in react-dom', async () => {
+    const client = createMockClient();
+
+    const { container } = render(
+      <UserDetailsWrapper client={client} suspend={suspend} />
+    );
+
+    if (suspend) {
```
umidbekkarimov: I kept using `toMatchInlineSnapshot` to keep to "visual" and without manual labor 🙂

src/__tests__/getMarkupFromTree-test.tsx
trojanowski: Good idea.

alidcastano:
@trojanowski when will a release with this PR be done? 

also, is server-rendering compatible with react-apollo, or will it only allow using one library?

trojanowski:
> when will a release with this PR be done?

@alidcastano I'm going to add documentation for this feature and then release it - I'll try to do it tomorrow.

> also, is server-rendering compatible with react-apollo, or will it only allow using one library?

Right now it's not compatible with react-apollo, but I'm going to add a compatibility layer.

trojanowski:
@alidcastano This PR is published as a part of release 0.3.0.

src/SuspenseSSR.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { Suspense, SuspenseProps, useContext } from 'react';
+import { SSRContext } from './internal/SSRContext';
+
+export function unstable_SuspenseSSR({ children, fallback }: SuspenseProps) {
+  const ssrManager = useContext(SSRContext);
+
+  return ssrManager ? (
+    <>{children}</>
```
sandiiarov: Was there a specific reason to wrap `children` into `fragment`?

PR #45, opened 2018-12-28
refactor(react-apollo-hooks): Move main modules to separate files.

umidbekkarimov:
Rollbacked changes from #39 for better git history.


src/utils.ts
```diff
@@ -1,5 +1,7 @@
 import isPlainObject from 'lodash/isPlainObject';
 
+export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+
 export default function objToKey(obj: Record<string, any>): null | string {
```
trojanowski: I think `objToKey` shouldn't be a default export here.
```suggestion
 export function objToKey(obj: Record<string, any>): null | string {
```

PR #46, opened 2018-12-28
refactor(tests): Use `toMatchInlineSnapshot` and `jest-react-profiler`.

umidbekkarimov:
Cherry picked changes from #42 related to test refactor.

Depends on #45.


src/__tests__/useQuery-test.tsx
```diff
@@ -101,310 +103,490 @@ const FILTERED_TASKS_QUERY = gql`
...
 
-it('should return the query data', async () => {
-  const client = createClient({ mocks: TASKS_MOCKS });
-  const { container } = render(
+function TasksWrapper({ client, ...props }: TasksWrapperProps) {
+  return (
     <ApolloProvider client={client}>
-      <Suspense fallback={<div>Loading</div>}>
-        <TasksLoader query={TASKS_QUERY} />
+      <Suspense fallback={<>Loading</>}>
```
trojanowski: I don't like `<Suspense>` is always used, even for test where `{ suspend: false }` is used. To fix that and minimize DRY we could:

1. Create a helper component:

```typescript
const SuspenseCompat: React.SFC<{}> = ({ children }) => <>{children}</>;
```

2. Add the following check inside `TasksWrapper`:

```typescript
  const SuspenseComponent = props.suspend !== false ? Suspense : SuspenseCompat;
```

3. Replace usage of suspense inside of `TasksWrapper` with `SuspenseComponent`.

trojanowski:
I added a few comments. I like the changes and LGTM except that.

PR #47, opened 2018-12-28
feat(useQuery): Use Apollo client state as main source of truth.

### LINKED ISSUE #29: Network error is not reported in errors and loading stays true

umidbekkarimov:
Closes #29.

Depends on #46.


umidbekkarimov:
@trojanowski ready for review

src/useQuery.ts
```diff
@@ -51,101 +49,115 @@ export interface QueryHookResult<TData, TVariables>
...
+      query,
+      variables,
+    }),
+    [
+      query,
+
+      pollInterval,
+      notifyOnNetworkStatusChange,
+
+      context,
```
trojanowski: Both `context` and `metadata` can be objects, so we should use `objToKey` also here. I think we should also change `objToKey` to correctly support all falsy values:

```typescript
export function objToKey(obj: Record<string, any>) {
  if (!obj) {
    return obj;
  }
  // ...
}

File: src/useQuery.ts [hunk 2]
```diff
@@ -51,101 +49,115 @@ export interface QueryHookResult<TData, TVariables>
...
   );
 
   useEffect(
     () => {
-      const subscription = observableQuery.current!.subscribe(nextResult => {
-        setResult(nextResult);
-      });
-      invalidateCachedObservableQuery(client, query, restOptions);
+      const subscription = observableQuery.subscribe(
+        () => setResponseId(x => x + 1),
```
trojanowski: I'd move it above to something like:

```typescript
const invalidateCurrentResult = () => setResponseId(x => x + 1);
```

src/__tests__/useQuery-test.tsx
```diff
@@ -461,18 +461,9 @@ it('should ignore apollo errors by default in non-suspense mode', async () => {
...
-  );
-  expect(consoleLogMock.mock.calls[0][1]).toMatchInlineSnapshot(
-    `"Network error: Simulating network error"`
-  );
-
-  // TODO: It should show error state.
   expect(container).toMatchInlineSnapshot(`
 <div>
-  Loading without suspense
+  Network error: Simulating network error
```
trojanowski: We should rename this test to e. g. _"shouldn't ignore apollo errors in non-suspense mode"_.

src/useQuery.ts
umidbekkarimov: I think we should encourage people to memoize these values by themselves, it's really easy to do with hooks: 
```js
const { accessToken } = useAuthPair();
const queryContext = useMemo(
  () => ({
    headers: {
      authorization: !accessToken ? undefined : `Bearer ${accessToken}`
    }
  }),
  [accessToken]
);
```

Also user can pass ANYTHING to this objects like regex or functions so `objToKey` would create false keys.

It's better to extend documentation about these properties.

trojanowski: I think we should provide good defaults for most cases. With the implementation from this PR is easy to miss that and in that case, no memoization will be used. In the current implementation on master [both `context` and `metadata` are serialized](https://github.com/trojanowski/react-apollo-hooks/blob/a303a6058ef1800ef360521f5798af3bfdef9e10/src/useQuery.ts#L75) (as a part of `restOptions`) and they still would be, even after merging this PR [in `queryCache`](https://github.com/trojanowski/react-apollo-hooks/blob/a303a6058ef1800ef360521f5798af3bfdef9e10/src/queryCache.ts#L50). We haven't had problems with that so far. We could tweak `objToKey` a little to better support non-objects (and also rename it) e. g.:

```typescript
export function optionsToKey(optionOrOptions: any): any {
  if (!isPlainObject(optionOrOptions)) {
    return optionOrOptions;
  }
  const sortedObj = Object.keys(optionOrOptions)
    .sort()
    .reduce((result: Record<string, any>, key) => {
      result[key] = optionsToKey(optionOrOptions[key]);
      return result;
    }, {});
  return JSON.stringify(sortedObj);
}
```

PR #50, opened 2018-12-31
docs: fix #49

### LINKED ISSUE #49: Invalid syntax in readme `Suspense` example
Regarding this snippet:
```js
import React, { Suspense } from 'react';

const MyComponent = () => {
  return (
    return (
      <Suspense fallback={<div>Loading...</div>}>
        <Dogs />
      </Suspense>
    )
  );
}
```
`return` statements are not valid in an expression context, and it seems like the example could be rewritten as:
```js
const MyComponent = () => (
  <Suspense fallback={<div>Loading...</div>}>
    <Dogs />
  </Suspense>
);
```

slikts:


README.md
```diff
@@ -112,15 +112,11 @@ To check if data is loaded use the
...
-const MyComponent = () => {
-  return (
-    return (
-      <Suspense fallback={<div>Loading...</div>}>
-        <Dogs />
-      </Suspense>
-    )
-  );
-}
+const MyComponent = () => (
```
trojanowski: Could you please change it to

```
const MyComponent = () => {
  return (
    <Suspense fallback={<div>Loading...</div>}>
      <Dogs />
    </Suspense>
  );
}
```

?

I haven't seen a component written like
```const MyComponent = () => (<SomeJsx />)```
in my codebase yet. :)

Comment (slikts) [reply README.md]: Sorry, I'm confused, because the very first example in the readme is this:
```js
const App = () => (
  <ApolloProvider client={client}>
    <MyRootComponent />
  </ApolloProvider>
);
```

trojanowski: Haha, good catch. Se let's leave it that way. I think I copied it from pupstagram example. Thanks :)

PR #62, opened 2019-01-18
chore(package.json): Add `sideEffects` entry.

umidbekkarimov:
See https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free


trojanowski:
Thanks @umidbekkarimov

PR #68, opened 2019-02-01
WIP: feat(useQuery): forward networkStatus

### LINKED ISSUE #67: how to check if fetching more, network status undefined
Using the query as 

```js
  const { data, loading, networkStatus, fetchMore } = useQuery(query, {
    variables,
    skip,
    notifyOnNetworkStatusChange: true
  });

```

`networkStatus` seems undefined when console logging so I can't do `const loadingMore = networkStatus === 3`

trojanowski:
Add `networkStatus` to `useQuery` result.
Closes #67

WIP: requires ugly hacks in suspense mode

### DIFF
--- a/src/__tests__/useQuery-test.tsx
+++ b/src/__tests__/useQuery-test.tsx
@@ -1,13 +1,13 @@
-import { ApolloClient } from 'apollo-client';
+import { ApolloClient, NetworkStatus } from 'apollo-client';
 import { ApolloLink, DocumentNode, Observable } from 'apollo-link';
 import gql from 'graphql-tag';
 import { withProfiler } from 'jest-react-profiler';
 import React, { Fragment, Suspense, SuspenseProps } from 'react';
-import { cleanup, render } from 'react-testing-library';
+import { cleanup, fireEvent, render } from 'react-testing-library';
 
 import { ApolloProvider, QueryHookOptions, useQuery } from '..';
 import createClient from '../__testutils__/createClient';
-import { SAMPLE_TASKS } from '../__testutils__/data';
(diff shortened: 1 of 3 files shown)


trojanowski:
Problem with the current implementation is that throwing `observableQuery.result()` subscribes to it (<https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/core/ObservableQuery.ts#L156>) and unsubscribes when the result is available. During the unsubscription the `isTornDown` flag is set (<https://github.com/apollographql/apollo-client/blob/cc3706ee76c0de7d56a7936a80ec6e6b96417520/packages/apollo-client/src/core/ObservableQuery.ts#L606>) which causes that the `NetworkStatus.error` is then returned as a value of `networkStatus` (<https://github.com/apollographql/apollo-client/blob/cc3706ee76c0de7d56a7936a80ec6e6b96417520/packages/apollo-client/src/core/ObservableQuery.ts#L186>). I was able to workaround it by creating an additional subscription (<https://github.com/trojanowski/react-apollo-hooks/blob/92c910fab4515ff8b15f75c7a5196fce58243080/src/useQuery.ts#L191>) which is unsubscribed after the next subscription in the `useEffect` hook is created (<https://github.com/trojanowski/react-apollo-hooks/blob/92c910fab4515ff8b15f75c7a5196fce58243080/src/useQuery.ts#L151>). Unfortunately it requires to use unstable API from the [scheduler](https://www.npmjs.com/package/scheduler) package.

trojanowski:
I'm closing it - `networkStatus` is already forwarded in the non-suspense mode. I'll try to add it to the suspense mode when Suspense for data fetching will be considered ready by the React team - maybe it'll be easier then.

PR #71, opened 2019-02-06
chore: Update React version to 16.8.0 🚀.

umidbekkarimov:
Most of unit tests start screaming:
```
      Warning: An update to Tasks inside a test was not wrapped in act(...).
      
      When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
      
      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://fb.me/react-wrap-tests-with-act
```
I found a solution for basic cases:
```js
jest.useFakeTimers();

function flushRequests() {
   // 1. Flush React effects
  flushEffects();

  // 2. Flush `apollo-mock` timers.
  act(() => {
    jest.runAllTimers();
  })

  // 3. Flush promises
  await new Promise(resolve => resolve())
}
```

But it doesn't work well in few scenarios where `apollo-client` updates content by itself.
E.g. `fetchPolicy: network-and-cache`, so we might have to enhance [invalidation behavior](https://github.com/trojanowski/react-apollo-hooks/blob/master/src/useQuery.ts#L144-L147).


trojanowski:
@umidbekkarimov I created an alternative implementation in #78. I was able to get rid all of those `act` warnings with this hack: https://github.com/trojanowski/react-apollo-hooks/blob/d4db5cef304a13a427f48f10cd322ac19cf307eb/src/useQuery.ts#L150-L157. Please let me know what do you think.

Issue #69, opened 2019-02-05
How to deal with suspend?


FredyC:
I am mostly curious what was the initial motivation to have `suspend = true` by default already. There is a lack of information on how to actually use suspense for data fetching considering that it's planned for late 2019. 

I just got burned by this badly when I forgot to set it false for one component deep down in the tree. I do have a `Suspense` way up in a tree to handle `React.lazy`. Suddenly when a query was rerun due to variables changing, everything got unmounted and fallback to top-level Suspense. It took me a while and some debugging to realize the culprit.

Do you have some actual examples of how to integrate it with react-apollo-hooks? Is it someone using it successfully already or is it everyone switching it off?

Let's consider a pretty simple scenario. A component that has `useQuery` and renders some markup with data from that query. Now with `loading` prop the decision what to show is done deeper in the tree. However, with suspension in play, it has to be done at top levels? I just cannot wrap my head around this for some reason.

trojanowski:
> I am mostly curious what was the initial motivation to have `suspend = true` by default already.

When I started implementation of this library there where no info when both hooks and suspense for data loading would be ready, but I thought that both of them could play nicely together. Right now we know that a stable release of React with hooks [should be available soon](https://github.com/facebook/react/pull/14692) and also it seems that the parts of suspense used by react-apollo-hooks [can be considered as safe to use](https://github.com/apollographql/react-apollo/issues/2539#issuecomment-459943586).

> I just got burned by this badly when I forgot to set it false for one component deep down in the tree. I do have a `Suspense` way up in a tree to handle `React.lazy`. Suddenly when a query was rerun due to variables changing, everything got unmounted and fallback to top-level Suspense. It took me a while and some debugging to realize the culprit.

I think it's a quite common use case. You'd like to show a loading indicator if there is no all data or assets required for the current page already available. With suspense you can have the whole logic for that in a single place, e. g:

```javascript
const MyLazyComponent = React.lazy(() => import("./AComponent"));

function ComponentWithData() {
  const { data } = useQuery(MY_QUERY);

  return <MyLazyComponent data={data} />;
}

ReactDOM.render(
  <Suspense
    fallback={
      <div>
        {"GraphQL data or lazy bundle loading (I don't care whitch one)"}
      </div>
    }
  >
    <ComponentWithData />
  </Suspense>
);
```

Without suspense, you'd have to support loading states separately in both `MyLazyComponent` and `ComponentWithData`.

[...]

PR #72, opened 2019-02-06
Change suspend to false by default

FredyC:
Based on discussion in #69


trojanowski:
@FredyC It's not such easy change. :) Tests are failing. If you don't have time for fixing them, I'll do it in the next few days.

FredyC:
Ah, right, haven't thought about that. Go ahead then, got too much on my plate already :)

trojanowski:
I created an alternative PR with tests and docs updated: #80

PR #78, opened 2019-02-11
chore: update React peer dependency to ^16.8.0 (alternative)

trojanowski:
BREAKING CHANGE: minimum supported (and tested) version of React is now 16.8.0

Closes #71

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
   "peerDependencies": {
     "apollo-client": "*",
     "graphql": "*",
-    "react": "^16.8.0-alpha.1"
+    "react": "^16.8.0"
   },
   "dependencies": {
     "lodash": "^4.17.11"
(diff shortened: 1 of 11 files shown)


umidbekkarimov:
TBH I don't really know what benefits we get from `act` (as far as I know we will need it for seamless transaction from class components to functional with hooks without breaking the test). So I guess shutting it down is ok. But on the other hand - wouldn't it be easier just to shut down `console.error`s with `act` messages ? :)

trojanowski:
> wouldn't it be easier just to shut down `console.error`s with `act` messages

I guess that using `act` can be helpful in some cases. We identified one place where it's not enough to get rid of warning messages. I think that maybe in the future those warnings can help us to identify some bugs, so it's a bad idea to silence them all.

PR #86, opened 2019-02-22
Updated README.md

nareshbhatia:
Returning a string a component creates the following TypeScript error in the parent component:

> Type error: JSX element type 'string | Element' is not a constructor function for JSX elements.
  Type 'string' is not assignable to type 'Element | null'.  TS2605

Better to return a proper element.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ const Dogs = () => {
     return <div>Loading...</div>;
   };
   if (error) {
-    return <div>Error! ${error.message}</div>;
+    return <div>Error! {error.message}</div>;
   };
 
   return (
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ const Dogs = () => {
     return <div>Loading...</div>;
   };
   if (error) {
-    return `Error! ${error.message}`;
+    return <div>Error! ${error.message}</div>;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -94,7 +94,7 @@ const Dogs = () => {
     return <div>Loading...</div>;
   };
   if (error) {
-    return `Error! ${error.message}`;
+    return <div>Error! ${error.message}</div>;
```
trojanowski: `$` is unnecessary here.

```suggestion
    return <div>Error! {error.message}</div>;
```

README.md
```diff
@@ -135,7 +135,7 @@ const GET_DOGS = gql`
 const Dogs = () => {
   const { data, error } = useQuery(GET_DOGS, { suspend: true });
   if (error) {
-    return `Error! ${error.message}`;
+    return <div>Error! ${error.message}</div>;
```
trojanowski: The same as above.

```suggestion
    return <div>Error! {error.message}</div>;
```

trojanowski:
Thanks @nareshbhatia

nareshbhatia:
Welcome! Sorry about the oversight on `$` signs!

PR #87, opened 2019-02-22
useQuery -> subscribeToMore

### LINKED ISSUE #14: Miss a lot of things? or it's me
Great work, but how I can access subscribeToMore on query hook and update the cache after a mutation with this hooks?

samdenty:
Closes #14. I've tested it and it works

Usage:

```ts
useEffect(() => query.subscribeToMore({ ... }), [])
```

### DIFF
--- a/src/useQuery.ts
+++ b/src/useQuery.ts
@@ -47,7 +47,7 @@ export interface QueryHookResult<TData, TVariables>
   extends QueryHookState<TData>,
     Pick<
       ObservableQuery<TData, TVariables>,
-      'refetch' | 'startPolling' | 'stopPolling' | 'updateQuery'
+      'refetch' | 'startPolling' | 'stopPolling' | 'updateQuery' | 'subscribeToMore'
     > {
   fetchMore<K extends keyof TVariables>(
     fetchMoreOptions: FetchMoreQueryOptions<TVariables, K> &
@@ -179,6 +179,7 @@ export function useQuery<TData = any, TVariables = OperationVariables>(
   ensureSupportedFetchPolicy(suspend, fetchPolicy);
 
   const helpers = {
+    subscribeToMore: observableQuery.subscribeToMore.bind(observableQuery),
     fetchMore: observableQuery.fetchMore.bind(observableQuery),
     refetch: observableQuery.refetch.bind(observableQuery),
     startPolling: observableQuery.startPolling.bind(observableQuery),


trojanowski:
`subscribeToMore` wasn't forwarded by purpose.  It's easy to make a mistake mixing with it with hooks. Please look at the discussion in https://github.com/trojanowski/react-apollo-hooks/pull/35. You can use `useSubscription` hook with the `onSubscriptionData` callback instead: <https://github.com/trojanowski/react-apollo-hooks#usesubscription>. Please let me know if you have any use case of `subscribeToMore` which is not possible to solve with `useSubscription`.

samdenty:
@trojanowski `useSubscription` looks good. Having `subscribeToMore` co-located on the query as well also does makes sense, as `fetchMore` is already there. Plus it's used in the Apollo documentation.

Side-note, `useSubscription` doesn't provide any way to unsubscribe

samdenty:
@trojanowski Now thinking about it, `subscribeToMore` is required. Hooks aren't conditional and always have to be called, meaning I can't delay the subscription.

Example:

```ts
function component() {
  const query = useQuery(...)

  // Can't accomplish this with useSubscription
  return <button onClick={() => query.subscribeToMore(...)}></button>
}
```

trojanowski:
> Side-note, `useSubscription` doesn't provide any way to unsubscribe

It's automatically unsubscribed when a component is unmounted. Or do you mean conditional subscriptions? For that I'm going to add `skip` option to it. Example usage:

```javascript
useSubscription(MY_SUBSCRIPTION, {
  // if `skip` is equal to `true` the subscription will be skipped
  skip: shouldSkip,
});
```

trojanowski:
> @trojanowski Now thinking about it, `subscribeToMore` is required. Hooks aren't conditional and always have to be called, meaning I can't delay the subscription.

`skip` option could be used for that. Your modified example:

```
function Component() {
  const query = useQuery(...)
  const [shouldSubscribe, setShouldSubscribe] = useState(false);
  useSubscription(MY_SUBSCRIPTION, {
    skip: !shouldSubscribe,
  });
  return <button onClick={() => setShouldSubscribe(true)}></button>
}
```

Please notice that in your example you don't unsubscribe - is it by purpose? If so you can always use `useApolloClient` and `client.subscribe`. If not - that's one of the errors I'm trying to prevent by not forwarding `subscribeToMore`.

samdenty:
@trojanowski skip option would also work albeit a bit messy, but it still wouldn't allow for dynamically created subscriptions:

```ts
function component() {
  cons[(clicks, setClicks)] = useState(0)
  const query = useQuery()

  // Each time the button is pressed, it creates a new subscription.
  return (
    <button
      onClick={() => {
        setClicks(clicks + 1)
        query.subscribeToMore({ variables: { clicks } })
      }}
    />
  )
}
```

> Please notice that in your example you don't unsubscribe - is it by purpose?

To highlight the above, but it'd make sense for subscribeToMore to automatically unsubscribe when the component unmounts

trojanowski:
@samdenty I thought about your use case, and I think it could be solved that way:

```typescript
import range from 'lodash/range';

function MyComponent() {
  const [clicks, setClicks] = useState(0)
  const query = useQuery()

  return (
    <>
      {range(clicks).map(clickId => (<Subscriber clickId={clickId} />))}
      <button
        onClick={() => {
          setClicks(clicks + 1)
          query.subscribeToMore({ variables: { clicks } })
        }}
      />
    </>
  )
}

function Subscriber({ clickId }) {
  useSubscription(MY_SUBSCRIPTION, { variables: { clicks: clickId }, });
  return null;
}
```

Please let me know if it works for you.

trojanowski:
@samdenty I'm going to close this PR right know. Please let me know if the solution from my previous comment doesn't work for you or if you have other use cases for `subscribeToMore` which are not possible to solve with `useSubscription`.

RodolfoSilva:
@trojanowski  how can I do that with `useSubscription` and `useQuery`:
```javascript
    subscribeToMore({
      document: newMessageSubscription,
      updateQuery: (prev, { subscriptionData }) => {
        if (!subscriptionData) {
          return prev;
        }

       // Perform updates on previousResult with subscriptionData
        return {
          ...prev,
          messages: [...prev.messages, subscriptionData.newMessage],
        };
      },
    });
```

https://www.apollographql.com/docs/react/api/react-apollo#graphql-query-data-subscribeToMore

PR #91, opened 2019-02-23
useQuery->stale

samdenty:
Apollo doesn't seem to provide the `.stale` property on the `currentResult`, and it also returns false for `loading` on the first `currentResult` loaded from cache.


trojanowski:
Hi @samdenty. What's your use case for this?

samdenty:
Stale returns `true` when the query was loaded from cache, and false after it's been fetched from the network

```ts
const query = useQuery(QUERY, { fetchPolicy: 'cache-and-network' })
return <div style={{ opacity: query.stale ? 0.5 : 1 }}>{query.data.messageCount}</div>
```

Issue #28, opened 2018-11-30
Rethrow apollo errors and recommend error boundaries


trojanowski:
Right now errors are returned as a part of the `useQuery` result. Maybe a better idea would be to throw them and recommend to use [Error Boundaries](https://reactjs.org/docs/error-boundaries.html#introducing-error-boundaries) instead (at least in the suspense mode).

So instead of this code:

```javascript
const Hello = () => {
  const { data, error } = useQuery(GET_HELLO);
  if (error) return `Error! ${error.message}`;

  return <p>{data.hello.world}</p>;
};
```

we could write:

```javascript
const Hello = () => {
  // we can be sure here that the error hasn't occurred
  return <p>{data.hello.world}</p>;
};
```

and use an error boundary at the root of the react tree (or anywhere else above the component) - similar to how we use the `<Suspense />` component.

fbartho:
Forcing an ErrorBoundary around every component seems painful at first glance. When `useQuery` I want to catch error-states. If you changed this, I think I'd have to do:

```javascript
let result
try {
  const {data} = useQuery(MY_QUERY);
  result = data;
} catch(e) {
  if (isPromise(e)) { throw e; }
  result = makeErrorData(e)
}
return (<Component {…result} />);
```

trojanowski:
I'd make it optional (but enabled by default) so if you wanted the old behavior you could use it like e. g. `const {data} = useQuery(MY_QUERY, { throw: false });`.

But the general idea is to make error management similar to what suspense does for data loading. So in most cases, it should be enough to put a single error boundary at the top of the React tree, but if you need it, you can also put it at the bottom, just above a component which requires a special behavior. Please compare:

```javascript
render(
  <Suspense fallback={<GlobalLoadingFallback />}>
    <MyPageComponent>
      <PageDataComponent />
      <Suspense fallback={<ImageLoadingFallback />}>
        <MyImageComponent />
      </Suspense>
    </MyPageComponent>
  </Suspense>
);
```

and:

```javascript
// a helper component but you can also easily implement it by yourself
import { ErrorBoundary } from 'react-error-boundary';

render(
  <ErrorBoundary FallbackComponent={<GlobalError />}>
    <MyPageComponent>
      <PageDataComponent />
      <ErrorBoundary FallbackComponent={<ImageError />}>
        <MyImageComponent />
      </ErrorBoundary>
    </MyPageComponent>
  </ErrorBoundary>
);
```

[...]

PR #93, opened 2019-02-24
BREAKING CHANGE: useMutation returns tuple with result

### LINKED ISSUE #90: Add second returned value to useMutation ?
WIth the Mutation component from Apollo. you can do something like this:

```jsx
const AddTodo = () => {
  let input;

  return (
    <Mutation mutation={ADD_TODO}>
      {(addTodo, { data }) => (
        <div>
          <form
            onSubmit={e => {
              e.preventDefault();
              addTodo({ variables: { type: input.value } });
              input.value = "";
            }}
          >
            <input
              ref={node => {
                input = node;
              }}
            />
            <button type="submit">Add Todo</button>
          </form>
        </div>
      )}
    </Mutation>
  );
};
```

could the useMutation hook be modified like so?

```jsx
const [doMutation, { data, loading, error }] = useMutation(...);
```

FredyC:
Closes #90 

The whole thing is sort of ripoff from Apollo's [Mutation component](https://github.com/apollographql/react-apollo/blob/master/src/Mutation.tsx#L173) converted to the Hooks. I've omitted a couple of things from there as those can be added later without breaking anything.

 * ignoreResults prop
 * onComplete & onError callbacks

---

I was thinking about this a lot. A major concern is how it fits into a whole Suspense & Error Boundaries (#28) paradigm. I mean it would be kinda nice to utilize that and have the same approach. However, mutation use cases are different than queries.

Considering simplest one...

```jsx
const MutantComponent = () => {
  const [mutateMe, { error, loading }] = useMutation(MUTANT)
  return (<div>
    <button disabled={loading} onClick={mutateMe}>I want to be a mutant</button>
    <ErrorDisplay error={error} />
  </div>
}
```

Since Suspense is not really ready, let's not worry about that (although it's a similar problem). But could we utilize Error Boundaries here? The error would be thrown within a `onClick` event handler and error boundary would do what? Throw away the component and replace it with some "there was error" and "retry" button? Doesn't feel like good UX. So I am kinda convinced that we need to be handling it on more granular level like I've drafted above.

@gaearon Would be lovely to get your opinion here if you can.


FredyC:
I just realized, since mutations are async, a thrown error cannot be caught by error boundaries 😢

One of the actual use cases where propagating error toward closest error boundary is displaying such error as a toast or even sending a report to some service. Such scenarios can be nicely generalized and component code gets a bit cleaner. 

So I've decided to implement an experimental approach which would throw the error synchronously by default. If anything, it gives much better error stack instead of just "unhandled rejection". And it can be switched off by `throwAsync` option if a need for a more granular error handling arises. Let me know what you think.

I've tested it by hand and it works nicely. Even without error boundary the error is thrown by React, it doesn't get lost anywhere. Only caveat is that whole component tree is thrown away in that case, but error clearly states to use error boundary, so I wouldn't be too worried about it.

FredyC:
@trojanowski I would love to hear an opinion from you if I should continue with adding tests/docs or it's something you don't want to merge at all.

trojanowski:
@FredyC the current form of `useMutation` was enough for me - I have usually used it in conjunction with [Formik](https://github.com/jaredpalmer/formik), where the loading and error state was stored. But I agree that there are many use cases where the implementation suggested by you makes perfect sense (maybe with the exception of the error handling). It would also make it easier to port code which used the `<Mutation />` component from react-apollo. Few comments:

> Since Suspense is not really ready, let's not worry about that (although it's a similar problem).

Actually, I can't see how Suspense would be helpful here.

> But could we utilize Error Boundaries here?

> So I've decided to implement an experimental approach which would throw the error synchronously by default. If anything, it gives much better error stack instead of just "unhandled rejection" inside of Apollo. And it can be switched off by `throwAsync` option if a need for a more granular error handling arises. Let me know what you think.

I don't think an error should be thrown at all by default. In my use cases, I needed access to the `error` as a part of the response, so I'd always had to disable this behavior. E.g., please look at the following code:

```
<Formik
  onSubmit={async (values, { setErrors, setSubmitting, setValues }) => {
    let response;
    try {
      response = await performMutation({ variables: { input: values }});
    } catch (error) {
      setErrors(getValidationErrors(error));
      NotificationManager.error('Cannot perform mutation');
      return;
    } finally {
      setSubmitting(false);
    }

    setValues(mapResponseToFormValues(response.data.myMutation));
    NotificationManager.success('Mutation completed');
  }
/>

```

Actually, I don't see a scenario where throwing them would be helpful. For cases where I think the rest of this PR would be useful (e. g. an easy possibility of re-running a failed mutation) it'd be better to not lost the state of the component which error boundaries would cause.

FredyC:
> Actually, I can't see how Suspense would be helpful here.

Yea, without a full picture how Suspense for data fetching will look like, I cannot imagine it either :) It pretty much boils down to eg. disable submit button which is very similar operation to showing spinner on data fetching.

> Actually, I don't see a scenario where throwing them would be helpful.

> > One of the actual use cases where propagating error toward closest error boundary is displaying such error as a toast or even sending a report to some service. Such scenarios can be nicely generalized and component code gets a bit cleaner.

Error boundary does not always mean you have to render some placeholder UI, it can be really just to collect errors and do something different with those. The challenge could be that the error boundary does not give much of context if it was thrown from mutation or a failed query which might require a different approach.

Anyway, if you look closer at [Mutation component](https://github.com/apollographql/react-apollo/blob/master/src/Mutation.tsx#L183), it already does throw by default, but since it's async code, it ends up as `unhandledRejection` which is not really helpful. I only turned it around so it throws synchronously, but can be turned off.

I guess Formik is not ready for this hooked world with such imperative API. If something like this would work, it would certainly clean up a lot of things. It's fairly contrived, but you get the gist :)

```jsx
const [performMutation, { data, loading, error }] = useMutation(...)
<Formik
  isSubmitting={loading}
  errors={getValidationErrors(error)}
  values={mapResponseToFormValues(data)}
  onSubmit={async (values) => {
    await performMutation({ variables: { input: values }});
    NotificationManager.success('Mutation completed');
  }}
/>
```

lotsoftick:
@FredyC thanks for your work. When are you going to release this ? :)

FredyC:
@lotsoftick I am kinda waiting for more opinions. There are still no tests and docs and I don't want to put an extra effort into before it's clear that it's wanted change.

lotsoftick:
@FredyC i think this is crucial, and must be done. 
```
const [mutateMe, { error, loading, data }] = useMutation(MUTANT)
```

FredyC:
Ok, I agree with that, but what about error handling part? Should it throw by default? The current behavior means you have to explicitly do this for every mutation call.

```jsx
const MutantComponent = () => {
  // using the new option here which represents current behavior
  const [mutateMe, { error }] = useMutation(MUTANT, { throwAsync: true })
  const mutateMeSafely = (...args) => {
    mutateMe(...args).catch(handleError)
    // or use async with try/catch which would way more verbose
  }
  return (<div>
    <button disabled={loading} onClick={mutateMeSafely}>I want to be a mutant</button>
    <ErrorDisplay error={error} />
  </div>
}
```

Compared to new default behavior from this PR the error is thrown synchronously which means it can be caught by error boundaries and sent to reporting service in a universal way without doing it every time the mutation is called. But it's still possible to grab the error from result and render it if that's what's required.

```jsx
const MutantComponent = () => {
  const [mutateMe, { error, loading }] = useMutation(MUTANT)
  return (<div>
    <button disabled={loading} onClick={mutateMe}>I want to be a mutant</button>
    <ErrorDisplay error={error} />
  </div>
}
```

Personally, I don't really mind which behavior goes by default as it's easy to wrap it into a custom hook. But I am trying to provide a best default experience.

FredyC:
The error boundaries still need a class component, but it's easy to have a single one like this sitting on top of the tree. 

```jsx
class ErrorBoundary extends React.Component {
  componentDidCatch(error, info) {
    // Report error, show the toast, whatever you like...
    logErrorToMyService(error, info);
  }

  render() {
    return this.props.children; 
  }
}
```

By directly rendering the children, no UI gets lots or remounted. It really just to catch that error in universal way. And best part is that you can make different customized error boundary, use it lower in the rendering tree to customize experience for a specific part of the app.

lotsoftick:
i prefer the first option, without Error Boundary

FredyC:
@lotsoftick Do you care to elaborate why? I am wondering if you are seeing some benefit there that I am missing.

lotsoftick:
mutation request may fail because of validation, let's say with error status 422 and with error messages  for each field in the body. I understand that this is can be done using ErrorBoundary as well, but basically my components look like this:

```javascript
const RoutComponent = ({ params, history }) => {
  const { loading, error, data } = useQuery(GET_LIST);

  return (
    <Fragment>
      {params.id &&
        <EditModal id={params.id} closeModal={() => history.goBack()} />
        /* show modal form over the table   */
      }
      {error && <div>{error.message}</div>}
      <table className={classnames({ loading })}>
        ...
        {data.map( /* render data */)}
        ...
      </table>
    </Fragment>
  )
}

const EditModal = ({ id, closeModal }) => {
  const { loading, error, data } = useQuery(GET_ITEM, { variables: { id }}); // getting item data
  const [mutator, { loading: updating, error: updateError }] = useMutation(MUTANT);

  const submit = async (e) => {
    await mutator(formData);
    closeModal();
  }

  return (
    <Modal>
      {error && <div>{error.message}</div>} {/* show error if can't fetch item data */}
      {updateError && updateError.status !== 422 && <div>{updateError.message}</div>} {/* show error if something went wrong during update. But error is not validation */}
      <ValidForm
        className={classnames({ loading: loading || updating })}
        onSubmit={submit}
        showValidationError={updateError.status === 422 ? updateError : null} {/* in case if data vas invalid show validation errors for each filed */}
      >
        {/* ... render inputs */}
      </ValidForm>
    </Modal>
  )
}
```

FredyC:
I see that I still haven't explained it well enough. You can do exactly that with this PR and default sync throw.

You will **always** get the error as the result from `useMutation`, that's for the fact and you cannot prevent that behavior.

Think of that sync throw as a bonus. A unified way of catching all mutation errors at one place (error boundary) and ideally reporting them to some service so nothing is going to escape your attention. Without it, you will only get `unhandledRejection` which you would need to catch manually for every single mutation call.

Is it better like this or am I really bad at explaining it? :)

lotsoftick:
@FredyC thanks, it's clear. Both will work for me, so, just use second one (
sync throw) as default :)

guilhermedecampo:
This is amazing, when are you able to merge @FredyC 	? 🚀

FredyC:
@guilhermedecampo Would love to get it out soon as well, but I am kinda fighting with tests.

Also @trojanowski doesn't seem to be convinced about it, so I hope he won't reject it later :)

trojanowski:
@FredyC 

> Error boundary does not always mean you have to render some placeholder UI, it can be really just to collect errors and do something different with those.

About collecting errors and sending them e.g. to Sentry - I don't think I'd like to send there any of errors being a result of invoking a mutation. They can belong to one of these groups

* network connection errors, e.g. if the user is online
* server-side resolver errors - they will be sent by a server Sentry integration
* [predefined Apollo errors](https://www.apollographql.com/docs/apollo-server/features/errors.html) - they are like 4xx HTTP errors

> The challenge could be that the error boundary does not give much of context if it was thrown from mutation or a failed query which might require a different approach.

I can see that you already tried to solve it in https://github.com/trojanowski/react-apollo-hooks/pull/93/commits/cf7b8b310af182fee37f52af3bb6ad5ad5058ed2. One suggestion - you can use an internal `Symbol` here as a marker.

>  Anyway, if you look closer at [Mutation component](https://github.com/apollographql/react-apollo/blob/master/src/Mutation.tsx#L183), it already does throw by default, but since it's async code, it ends up as `unhandledRejection` which is not really helpful.

It just simulates the behavior of `apolloClient.mutate()`. It's just like:

```typescript
try {
  const result = await client.mutate(...);
  doSomethingWithResult(result);
} catch(error) {
  throw error;
}
```

`unhandledRejection` can also be sent to an error collecting service, but as I said above, I don't think it's a good idea. In my opinion async errors works best if you try/catch them manually, like in the Formik example in my previous comment.

My idea is to replace `throwAsync` option with something like `throwMode`. It could have 3 possible values:

* `sync` - the behavior in the current version of react-apollo-hooks. It works well in almost all cases for me. I think it should be the default value.
* `none` - don't throw errors at all, just set the value of the `error` state. That way we'd have something like an "internal" error boundary. We could check the presence of the `error` and, e.g. show "Try again" button.
* `async` - the default behavior in the current version of this PR. I actually don't see too much use cases for it (which wouldn't be solved by the `none` throw mode).

> @trojanowski doesn't seem to be convinced about it, so I hope he won't reject it later :)

I like the API. I'm just not sure about the error handling part.

@umidbekkarimov @rosskevin what's your opinion about it?

FredyC:
> About collecting errors and sending them e.g. to Sentry - I don't think I'd like to send there any of errors being a result of invoking a mutation. They can belong to one of these groups

> * server-side resolver errors - they will be sent by a server Sentry integration

Yea, I do care about those errors and I want them in client-side Sentry simply because there might some relevant app state that can help figure out an issue. Perhaps I've sent wrong variables and I would like to know why and how it happened.

> I can see that you already tried to solve it in [cf7b8b3](https://github.com/trojanowski/react-apollo-hooks/commit/cf7b8b310af182fee37f52af3bb6ad5ad5058ed2). One suggestion - you can use an internal `Symbol` here as a marker.

I am not sure if there some real benefit to it. I got inspired by ApolloError and how is the `isApolloError` implemented :)

The `throwMode.none` feels pretty much useless because you cannot render any UI from the hook. Someone might misunderstand, use it and the errors would be suddenly lost unless they manually check the `error` in every mutation call. It's fairly dangerous mode to have there. I would rather not.

---

I have a slightly different idea. We can have `sync` mode by default, but we can export something like this from the package.

```jsx
const throwModeContext = React.createContext()

class ErrorBoundary extends React.Component {
  componentDidCatch(error, info) {
    console.error(error)
  }

  render() {
    return <throwModeContext.Provider value="async">{this.props.children}</throwModeContext.Provider>; 
  }
}
```

This way it's possible switch the mode not globally, but for that part of tree where we actually know there is an error boundary to catch them. It's much safer. The context Provider itself could be exported as well for a customized error boundaries.

This gives a great power as you can keep part of the tree to handles sync errors and other part would be throwing async and catching with boundary. Most importantly, you don't need to think about it. Just drop boudary where you want and rest of the app will be safe.

umidbekkarimov:
> @umidbekkarimov @rosskevin what's your opinion about it?

@trojanowski I'm good with this proposal as long as we keep current hook functionality (e.g: as  `useMutationFn`).

TBH I'm not using `useMutation` directly in my project, I merged it with `react-final-form-hooks` as `useMutationForm` so `final-form` deals with all state management for me.

___

But I'm not sure about using context to manage hook options, It makes it counter intuitive. It's still better to wrap with custom hook to keep it straightforward.

FredyC:
> But I'm not sure about using context to manage hook options, It makes it counter intuitive. It's still better to wrap with custom hook to keep it straightforward.

It's not meant like that. You would still be able to override that `throwMode` per `useMutation` meaning you can surely wrap it into a custom hook. The idea is merely to have an easy escape route with a primitive error boundary to never lose any error and have useful information about it. The `sync` mode which yields `unhandledRejection` gives absolutely no clue of where the error happened and that's the major reason why I don't like it.

sami616:
this looks 😍merge merge! 😁

rovansteen:
I’m just wondering what the appropriate way to handle these errors in the error boundaries would be. Right now I use error boundaries to crash my application by throwing errors I can’t recover from. With this change I would need to know that the error is from a mutation (for example a form) and we can safely render the children and display the error with some sort of an alert in the form.

FredyC:
@rovansteen Yea, I haven't mentioned it here just yet, but the last commit contains some utility for that. I am planning to add the same thing for queries, later on, to be able to define opposite logic if necessary.

```jsx
import { isMutationError } from 'react-apollo-hooks'
class ErrorBoundary extends React.Component {
  state = { bailout: false }
  componentDidCatch(error, info) {
    if (!isMutationError(error)) {
      this.setState({ bailout: true })
    }
    console.error(error)
  }

  render() {
    if (this.state.bailout) {
      return <div>It has crashed</div>
    }
    return this.props.children
  }
}
```

trojanowski:
> I'm good with this proposal as long as we keep current hook functionality (e.g: as useMutationFn).

@umidbekkarimov actually you could quite easily get the current functionality. It would be enough to replace the following invocation:

```typescript
const mutate = useMutation(MY_MUTATION);
```

with that one:

```typescript
const [mutate] = useMutation(MY_MUTATION);
```

Would it work for you? I'd prefer not to add a separate hook just for that. @leonardfactory @DAB0mB @Urigo Is that change OK for you? Would it be problematic for graphql-code-generator?

> TBH I'm not using useMutation directly in my project, I merged it with react-final-form-hooks as useMutationForm so final-form deals with all state management for me.

Good to know it's not only me using mutations with a form library. :)

> The throwMode.none feels pretty much useless because you cannot render any UI from the hook. Someone might misunderstand, use it and the errors would be suddenly lost unless they manually check the error in every mutation call. It's fairly dangerous mode to have there. I would rather not.

I thought it's the biggest benefit of this PR. :) You cannot render any UI from the hook, but you can use the presence of an error to decide what to render in your component. So if you wanted such behavior and didn't want to send those errors to a collecting service you would have to use the hook that way:

```typescript
const [mutateBase, { error }] = useMutation(MY_MUTATION);

async function mutate(options) {
  try {
    await mutateBase(options);
  } catch (errorToIgnore) {
    // ignore the thrown error to prevent `unhandledRejection` event
    // we use `error` returned by `useMutation` to decide what to render
  }
}
``` 

Anyway, I have an alternative suggestion. Let's forget about `throwMode` and introduce a different option - `errorHandler`. It would be a function run when an error happened. The default implementation would be:

```typescript
const defaultErrorHandler = error => { throw error; }
```

So it would work the same as the current version of `useMutation`. But you can replace it e.g. with `noop` (a more explicit version of the `none` throw mode), or a handler sending an error to Sentry:

```typescript
function useSetryFriendlyMutation(mutation, options) {
  return useMutation(mutation, {
    ...options,
    errorHandler: error => {
      Sentry.captureException(error);
    }
  );
}
```

That way you don't need to use `isApolloError` or `isMutationError`. But if you still prefer to use error boundaries you can easily create a custom hook for that:

```typescript
function useMutationWithErrorBoundariesAnyway(mutation, options) {
  const [mutate, result] = useMutation(mutation, options);
  if (result.error) {
    throw result.error;
  }
  return [mutate, result];
}
```

> But I'm not sure about using context to manage hook options, It makes it counter intuitive. It's still better to wrap with custom hook to keep it straightforward.

I agree. But if still prefer such behavior you can do it easily with a custom hook and a context provider.

FredyC:
> But if you still prefer to use error boundaries you can easily create a custom hook for that:

Error boundaries are becoming fairly common practice. I don't like that you would not have an option to use them directly and everyone would have to enable it with a custom hook. I would be really careful about building library in a way that needs custom hooks to work properly.

> Let's forget about `throwMode` and introduce a different option - `errorHandler`. It would be a function run when an error happened. The default implementation would be:

I don't follow how is that good thing. In most cases handling an error does mean you want to log/report it somewhere. It would be super annoying to need to do it everywhere. I like my DRY code :)

I am starting to like `throwMode`. And I don't mind having the `none` option after all (it already helped with tests). It can come in handy in scenarios where you really do want to just render error in some way (eg. icon), but you don't want to report it anywhere. It just needs proper warning in the documentation. I am thinking about a separate document about error handling in general.

And I made a wrong statement above. I've meant to keep `async` mode as default. I am perfectly fine with that. Then the error boundary can automatically switch it to `sync` for a given subtree so those errors can actually bubble up.

leonardfactory:
@trojanowski 

> Would it work for you? I'd prefer not to add a separate hook just for that. @leonardfactory @DAB0mB @Urigo Is that change OK for you? Would it be problematic for graphql-code-generator?

Seems like it would be supported by default. Since we just return `ReactApolloHooks.useMutate(...)` result entirely, it should work by default! I'll make sure however anything plays well if and when this will be released

> Good to know it's not only me using mutations with a form library. :)

+1 here 😄 Furthermore, Formik is currently [working on v2 version based on hooks](https://github.com/jaredpalmer/formik/issues/828)

FredyC:
Alright, it's time ❗️ I am opening PR for a review as I have finally managed to figure tests. I will work on the documentation in the meantime.

codingwaysarg:
Great guys! I didn't start using this library because of this missing feature. I'm excited about this PR! Thank you so much.

src/useMutation.ts
```diff
@@ -32,18 +37,155 @@ export type MutationFn<TData, TVariables> = (
...
+      const mutationId = generateNewMutationId();
+
+      try {
+        // merge together variables from baseOptions (if specified)
+        // and the execution
+        const mutateVariables = options.variables
+          ? { ...mutateOptions.variables, ...options.variables }
+          : mutateOptions.variables;
+
+        const response = await client.mutate({
```
umidbekkarimov: I would hardly recommend not to use `async-await` syntax. It increases amount of [output code](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=KYDwDg9gTgLgBAMwK4DsDGMCWEVyQZ2AFkkYBDLHACgCg44BbUi7FAGjrgCMzCB5MJRT44AXjgBvAL40AlJM5oc-eGgA2mYCnjiCwAIKQ1aiAGENWmFR79BrfADp1m7bIDci5fADaUYPiQ1GDY4QhgAJX9AmABdMThIsgwHPQBlchhgWnp6AHNgGABJFEwsMjV0imBOdxpPYXgJOBgACygIAHciCAATYHiAcl4AT3QBkIdJiDtlOCl4m2ABIXwPTkTkvQBRBARgDGycqnlRAD4FHJzMBDgqVvau3v7RF7gB_FG0AbgAMh-4PwBIIOYBQdpQeQSTiXej3ToAqLA0HgjwwubQuYcS7eOGPPohQHREFg6AxGpregAekpCLCoQy_WucGcljgmAaZHQ_TQLU5-Xw62ASRgKUIABkyMMIKQdnsDhjjmJzlC0WFIkCrPkiiUyhUGbVLlIsTlvCztGT6LV6ipJHB8ihQVUAHLALrMISFHohTD4boqSJoSwkDKsObxPTBlg4AAqUCSAGt2bljhTmV44DhI0JKrB4oqzhcrjcqABCQnAkxkHpJyEYqk0_RwXloePNCB2goAzA9BJCjBwDpkKAlFC5EQ9JBQJPNfwwRx1uDCgASCao-eVC_oasRVhVaJyaHKamAPQAXM0oEhgMb93AehQyOfUH0EOyTzf98joE-UC-317NxyXl8C2EkoHPBBykID80UratR3PGBL2vQCpANGE0NRHIZHoKRUyUBoMxQLNWFA8E8y_KAQiYEMcE9E4N0uJkqB9P0In2IN3VYKgaKjFB6NrNFqTgRtm1bGB2y1Lsew2eBB2HJNx0nadMhUec0WXVd10LVUCnVaIqDAPwADclVuPdb0mBwjOAYyYJhSjzwdDoRKMEwSJwMjoCoSjqK4nAQkWZZ7FkezLmArzwIvK8wpyOCkwgqCUNvND0MNNL0Vw_D00zfyUFMCAGDAY9Mh7cRDnoQFIGEZKcl4j0ehqMyLLTQimnvcgQkokR5nEKrlGALD6GY7rfn-bqHGPUdWjgc4AAZBJhXLaJQSKqGc1yIGMCA1qaXI4zAFoAEUxUi_Bz1GtC_JW-ihpyPwYEnFA7pkDFmNYiB_Q47QPJQHi8oEnScmE0SEzbDt4CnGS-zkocRzHO9lNHGc1IcBdNJbNcGKBy5tw1QyTLMqgWrRKybLswD6A6shYvoeKEMQJLYtSu7cIynC5mywjL2IvL4lk0VgFMI8eExjERnQW56qWGZhGxkmiN-nMrAy-gCJterWE9eJ7UdTIXTdG6ehTOo0SQ4YceBmkGFBfI2y1FpQTgYyh0wMguGPEQEHaBhuF4GWVluJl8DAfZrk0HpZEA4TOR7Vp-lAfZSFYQD1fgaWADVXfdz34mmFYHBdqcc_8Sm4AAfltKzpeC5RC-zj3_AmKZZccIu3cbnqy_PGvW_r4vO9TNE09parCHiMhB1KZkLG0BxpeJsvNYCsurPz-xabgav3QDjey_bkvzsYHes4Hz3matW9lr4gqipKk8qH6mrrr427AIep7R4G1mw0PGAeVuMiRaaJr5CDWsiF-DVVZFluLibofQxCvCGB8MYwD9y4jgMiH-uF34FE_tIH-HNcIflNLPYIx8VqBX9rXYQIQIBcAAFbRggAAaWAMMaw1DW6yAtHAS-CJHrDjgL4VAv0CQ7hiB4GQQA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=false&fileSize=true&timeTravel=false&sourceType=module&lineWrap=false&presets=react%2Ctypescript%2Cenv&prettier=false&targets=&version=7.3.4), makes it hard to debug in browser devtools and requires `regeneratorRuntime` polyfill.

FredyC: Thanks, async-await is a goner :)

FredyC:
@trojanowski Just tried to actually build this and it's failing because of class properties used in those Error classes. Would you be against adding https://babeljs.io/docs/en/babel-plugin-proposal-class-properties?

medson10:
Amazing job

FredyC:
So I am trying to implement this change into our production app and I am having some second thoughts about error handling.

First and foremost, error boundaries for mutations are a bad idea. See my [post at Spectrum](https://spectrum.chat/react/general/error-boundaries-as-unified-handler-or-i-thought~ddc764e4-598f-475c-89d0-d3591560d40e) for details. This is something I haven't realized before so I am going to drop the `throwMode` from the PR after all.

Also the `none` for the `throwMode` does make sense now. Consider use case like following (we have bunch of those).

```js
const [zoneUpdate, { called, data, error }] = useMZoneUpdate({ throwMode: 'none' })

if (called) {
  if (data) {
    routeToZoneList()
    enqueueSuccess(`Zone has been updated`) // show the toast
  }
  if (error) {
    enqueueError(`Zone failed to update`) // simplified
  }
}

const onSubmit = values => {
  zoneUpdate(zoneModel.getZoneVariables(zoneId, values))
}
```

I kinda like this more instead of `try-catch` and awaiting in every single case. That error handling can be unified with a custom hook and basically send it to the Context to show the relevant toast with errors from the backend server.

If there would be an error thrown in this use case, it ends up as `unhandledRejection`. And I don't want to `try/catch` it in every single component. So the option for not throwing at all definitely makes sense.

---

Stay tuned for more discoveries, but this PR is far from ready. Feel free to use this in your project as well to figure out some best practices. It's pretty much just copy & paste one file, you don't need to wait for official release.

gijo-varghese:
Right now I'm using this library for queries and 'react-apollo' for mutations. I'd love to see this pr merged soon. 

Also, regarding Error boundary I think it's a bad idea. I'd like it work exactly like how 'react-apollo' works. If there is an error I'll show another div based on it

@FredyC can I expect it soon?

trojanowski:
@FredyC 

> First and foremost, error boundaries for mutations are a bad idea. See my post at Spectrum for details. This is something I haven't realized before so I am going to drop the throwMode from the PR after all.

👍. BTW I did some testing and I think that the unhandled rejections are not so bad for error reporting - you can read a meaningful stack trace from the [`unhandledrejection` event](https://developer.mozilla.org/en-US/docs/Web/Events/unhandledrejection), and sentry supports it, but unfortunately, it requires polyfills on Firefox/Edge and is not supported on IE.

> First and foremost, error boundaries for mutations are a bad idea. See my post at Spectrum for details. This is something I haven't realized before so I am going to drop the throwMode from the PR after all.
> Also the none for the throwMode does make sense now.

So we could abandon the `throwMode` option and introduce an alternative one, e.g. `rethrow` which would control that behavior (`true` would mean the `async` throw mode described in <https://github.com/trojanowski/react-apollo-hooks/pull/93#issuecomment-468671007> would be used, `false` would enable the `none` mode, `sync` wouldn't be implemented at all). I prefer it to be `true` by default.

> Stay tuned for more discoveries, but this PR is far from ready.

What other changes are you considering?

FredyC:
@trojanowski I've just pushed an update. Documentation is still lacking in my opinion, but otherwise it seems to be working. I've included `rethrow` the way you said, agree on that. I've also made `ApolloOperationError` which is universal for all three types of operation and comes handy when recognizing errors in some unified error handler.

> What other changes are you considering?

I am not sure yet. Trying to use this in as many cases as I can, but I am also a bit swamped with some tasks.

> but unfortunately, it requires polyfills on Firefox/Edge and is not supported on IE.

Yea, one of the reasons why I wouldn't use it. Also, catching it is feels rather hacky.

---

For my own use I've figured out a nice way with Context as a universal error handler. It basically looks like this and it seems to be working quite nicely so far. 

```tsx
export const DefaultErrorHandler: React.FC = ({ children }) => {
  const onRetry = React.useCallback(() => {
    // reset error state
  }, [])

  const onUnhandledError = React.useCallback((error: Error) => {
  }, [])

  const onNetworkError = React.useCallback((error: Error) => {
  ), [])

  const onOperationError = React.useCallback(() => {
  }, [])

  const onUserErrors = React.useCallback((messages: ReadonlyArray<string>) => {
  }, [])

  const render = () => {
     // render fallback UI based on kind of error
     // or keep the current one - without remounting!
     return children
  }

  return (
    <ApolloDefender
      onNetworkError={onNetworkError}
      onUserErrors={onUserErrors}
      onOperationError={onOperationError}
    >
      <SillyErrorBoundary onError={onUnhandledError}>
        {render()}
      </SillyErrorBoundary>
    </ApolloDefender>
  )
}
```

The `SillyErrorBoundary` is just a `componentDidCatch` passing error in here (it's included in the PR). The `ApolloDefender` is my opinionated Sentry centered component which exposes `onError` through the Context which can be used for passing any caught errors (queries, mutations, subscriptions).

gijo-varghese:
Tests look good. Are we good to merge this?

FredyC:
> Tests look good. Are we good to merge this?

I think the docs really need more information, but I am a bit swamped this week. Could use some help with that really. Or we can merge this now and leave docs for a separate PR so someone else can do that. What do you think @trojanowski?

gijo-varghese:
I see that basic doc is already done. It would be if you guys can merge this asap and create a separate PR for docs later. I can also help with the docs

FredyC:
@trojanowski So can you actually merge it? I don't think there is going to be any other breaking change addition right now.

trojanowski:
@FredyC I'd prefer to reduce this PR and remove _SillyErrorBoundary_ and _ApolloOperationError_. I think _SillyErrorBoundary_ in the current form is not very useful -  a better idea is to document error handling. Also, _ApolloError_ seems enough for me (instead of _ApolloOperationError_) - in your `onError` handler you know that the error was thrown by a mutation (and you can determine by which one).

FredyC:
> I think _SillyErrorBoundary_ in the current form is not very useful

That depends, in React Hooks world it's very useful as you don't have a hook variant (yet). It's a useful workaround so you don't need to copy-paste it in your code. I agree it doesn't really fit well in the package, but I have to keep in the code for the tests. So why not reexport when it's already there? :)

> Also, _ApolloError_ seems enough for me (instead of _ApolloOperationError_) - in your `onError` handler you know that the error was thrown by a mutation (and you can determine by which one).

How can you know which mutation has thrown in a universal error boundary? Or if it was a mutation at all. That's kinda point of that Error class so you can unify your error handling into a single point and do a decision what to do based on the operation failing.

joenoon:
Getting up to speed on this and I've picked up a couple nice patterns from the comments that beat the fully imperative routine of try/catch/handle I usually do, so thank you.  I think this reduces the code needed around calling a mutation - considerably if you only need to display the loading/data/error declaratively.  In my case I usually need to do something with the response imperatively, but even then this still reduces loading and error and the need to catch.  I think I like it!

gijo-varghese:
Looks like official `react-apollo` is getting ready for hooks! https://github.com/apollographql/react-apollo/pull/2892

FredyC:
@gijo-varghese That won't be ready for some time yet, there was a word they want to wait for Suspense for data fetching which might be by the end of the year.

gijo-varghese:
@FredyC can you merge this soon?

FredyC:
@gijo-varghese Well, it's not up to me really, @trojanowski had some objections, I've responded. If you are impatient, just grab the [Yalc](https://github.com/whitecolor/yalc) and you can roll without waiting. That's what I do.

trojanowski:
> > I think _SillyErrorBoundary_ in the current form is not very useful
> 
> That depends, in React Hooks world it's very useful as you don't have a hook variant (yet). It's a useful workaround so you don't need to copy-paste it in your code. I agree it doesn't really fit well in the package, but I have to keep in the code for the tests. So why not export when it's already there? :)

Because it would be a part of the public API and we'd have to support it. Also, I'd instead prefer to recommend a specialized library for that, like [react-error-boundary](https://github.com/bvaughn/react-error-boundary). Moreover, I don't think `SillyErrorBoundary` is even useful for tests in this PR - you use it to check if no errors were throws while rendering the React tree. However, you don't need an error boundary for that - tests throws by default in that case.

> > Also, _ApolloError_ seems enough for me (instead of _ApolloOperationError_) - in your `onError` handler you know that the error was thrown by a mutation (and you can determine by which one).
> 
> How can you know which mutation has thrown in a universal error boundary? Or if it was a mutation at all? That's kinda point of that _ApolloOperationError_ so you can unify your error handling into a single point and do a decision what to do based on the operation failing. When this is merged, I would like to introduce the same thing into `useQuery` and `useSubscription`.

But we decided that [we won't support the synchronous error throwing](https://github.com/trojanowski/react-apollo-hooks/pull/93#issuecomment-473656506), so in order to use error boundaries (or sending it to the context like in your comment) you'd need a custom hook - you could convert it to a "higher order error" there. We could continue the discussion about `ApolloOperationError`, but I'd prefer to do that in a separate issue or PR - if we'd merge this PR in the current form, some public API functions would behave incorrectly (`isQueryError`, `isSubscriptionError`).

FredyC:
> But we decided that [we won't support the synchronous error throwing](https://github.com/trojanowski/react-apollo-hooks/pull/93#issuecomment-473656506), so in order to use error boundaries (or sending it to the context like in your comment) you'd need a custom hook

It has nothing to do with the synchronous throw. It's about an error object in general which should carry that information to improve debugging. I have this in my wrapper which allows me to pass an error to closest Context (ApolloDefender) which can decide what to do based on information in that error. If it's just regular ApolloError, that information is lost.

```js
  const { onError } = useApolloDefender()
  React.useEffect(
    () => {
      if (result.called && result.error) {
        onError(result.error)
      }
    },
    [result.called, result.error],
  )
```

I think I am going to hand this over to you and do changes you deem necessary. We obviously have different opinions about it and nobody else seem to care enough. I am already using my custom fork of this library so I don't really care that much to get this merged.

charlie632:
When is this getting merged? I need error checking for a crucial part in my application flow. Is any workaround for this?

guilhermedecampo:
@charlie632 http://bfy.tw/N54W

sami616:
Hi folks, just checking if this is close to being merged? really looking forward to it!

FredyC:
Perhaps, instead of asking for a merge to be done soon could actually contribute with an opinion on standing questions. It would ultimately help much more than poking and prodding like impatient nerds. 🙊

trojanowski:
Next week I'm going to try to add changes I requested in https://github.com/trojanowski/react-apollo-hooks/pull/93#issuecomment-479462989 and merge after that.

sami616:
@trojanowski thanks for the insightful response.

groundmuffin:
@FredyC, I appreciate your efforts

jeffreyffs:
> Edit: I am currently working on an article showing that error boundary is fairly useless in its current form and Context has surpassed it.

@FredyC I'm super curious on this topic and was wondering how this article is coming along? Would love to see your implementation of a context based error handler :)

trojanowski:
I added an alternative PR with changes I requested in my previous comment: #135. @FredyC please let me know what do you think about it.

FredyC:
@trojanowski Well, you know my opinion. I could live with the removal of SIllyErrorBoundary, but those errors wrapped are essential. I am already using those in production and it has proven very valuable in automated bug reports. Shame nobody else sees it. So be it, I will stick to my version.

gijo-varghese:
Finally! `npm install @apollo/react-hooks@beta --save`

Soon I'll run `npm r react-apollo-hooks`

https://twitter.com/peggyrayzis/status/1128280687470100480

ctrlplusb:
I understand you are excited about this feature @gijo-varghese - but responding like this sounds quite insensitive to the fact that a lot of people donated their time and effort to make this library freely available to you. Fair enough if there is an alternative but please show some a moderate amount of appreciation for what has been gifted.

maxschmeling:
This library has been incredibly helpful and well done. To give @gijo-varghese the benefit of the doubt, I hope they just thought that it was a transition to Apollo ownership / maintenance and they were genuinely excited for the library.

PR #95, opened 2019-02-25
feat(useQuery): accept client option

### LINKED ISSUE #85: Consider allowing `client` to be passed as an argument.
The [error message](https://github.com/trojanowski/react-apollo-hooks/blob/75fbecd5004e606e8f6c8868f8dd1e23a3cca57b/src/ApolloContext.tsx#L27-L30) from `useApolloClient` suggests that it should be possible to pass the `ApolloClient` as a prop.  While it's just the error message from `react-apollo` verbatim, it still seems like a valid enough option.

alexkirsz:
Fixes #85.

I chose to preserve hooks invariants by still calling the `useApolloClient` hook despite the Apollo client being passed from the options.

I also implemented the same behaviour for `useMutation` and `useSubscription`.

This allows for creating custom hooks that address a specific client, e.g.:

```ts
function useBooksQuery(query, options) {
  const booksClient = useContext(BooksClientContext);
  return useQuery(query, { client: booksClient, ...options });
}
```

### DIFF
--- a/src/ApolloContext.tsx
+++ b/src/ApolloContext.tsx
@@ -19,8 +19,17 @@ export function ApolloProvider<TCacheShape = any>({
   );
 }
 
-export function useApolloClient<TCache = object>(): ApolloClient<TCache> {
+export function useApolloClient<TCache = object>(
+  overrideClient?: ApolloClient<TCache>
+): ApolloClient<TCache> {
   const client = useContext(ApolloContext);
+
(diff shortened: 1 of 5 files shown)


jdenderplays:
This would be extremely useful for multiple clients instead of having to mess around with the context.

trojanowski:
Thanks @alexkirsz. Published with release v0.4.3.

PR #98, opened 2019-02-27
feat(useSubscription): Add skip option

### LINKED ISSUE #97: Skip subscription
Given that hooks cannot be used conditionally, it makes sense to have `skip` option for subscriptions too. A use case is when for example necessary variables are not available yet because the user needs to set the filter first.

FredyC:
Closes #97 

Fairly primitive change in the essence :)


trojanowski:
Thanks @FredyC. I just published v0.4.2 which has this PR included.

Issue #99, opened 2019-03-02
useQuery should return cached data after error


madflanderz:
When using fetchPolicy **cache-and-network** and then happens an error useQuery returns with empty data. I would expect that the previous data is returned. React-Apollo is doing this as well, you can see this here: https://github.com/apollographql/react-apollo/blob/master/src/Query.tsx#L444

Steps to reproduce:

- show component that uses useQuery with  fetchPolicy="cache-and-network", data is received
- disable network (or block graphql server request in dev tools)
- show component again, useQuery is executed again but fails with network error, data object is empty

Expected behaviour:
useQuery should return previous cached data when error happens.

I already tried the solution and it seems to work fine. What do you think about this?
https://github.com/madflanderz/react-apollo-hooks/commit/cad5f7525d93ad2d1f54df097623b4767fdfe89d

pontusab:
Related? https://github.com/trojanowski/react-apollo-hooks/issues/74

madflanderz:
No, the title looks similar but #74 is another problem. The usecase for this problem here is this:

- you have 2 pages in your app that each use useQuery with fetchPolicy="cache-and-network"
- visit both pages, user can see data from first requests
- turn of internet 
- visit both pages again, user should still use the app and see the data from the cache

pontusab:
@madflanderz nice 👍

trojanowski:
Should be fixed in v0.4.4.

PR #100, opened 2019-03-02
useQuery returns cached data when there is an error

madflanderz:
Related Issue with all Info: https://github.com/trojanowski/react-apollo-hooks/issues/99

Maybe would be nice to write a test for this.

### DIFF
--- a/src/useQuery.ts
+++ b/src/useQuery.ts
@@ -135,8 +135,17 @@ export function useQuery<
     () => {
       const result = observableQuery.currentResult();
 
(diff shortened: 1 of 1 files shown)


fbartho:
@trojanowski this seems super useful. Given that Apollo seems (to me) like it's dragging its feet on directly integrating their own hooks, do you think we could get this and other fixes merged?

trojanowski:
@madflanderz Thanks and sorry for the late reply.

trojanowski:
Published as v0.4.4.

Issue #84, opened 2019-02-15
Test runner throws act() warning on waitForNextTick - discussing better solution for React 16.8.1


mateusznowosielski:
After upgrading to React 16.8.1 (react-testing-library@5.8.0), tests throw warnings whenever we use `waitForNextTick` with Promise and setTimeout, which is everywhere when we need to wait for a component to rerender with result data.
```
act(() => {
    /* fire events that update state */
});
```
Sample test: https://github.com/trojanowski/react-apollo-hooks-sample-test/blob/master/src/__tests__/Hello-test.js 

As this is just a temporary solution to the problem with useEffect hook
```
render(null);
await waitForNextTick();
```
I came up with below solution which is far from pretty and elegant, but in simple scenario it works. Since `act` callback can't return anything, as for now it's not supported, I had to wrap the above code inside another self executed async function.
```
// Mocking queries and rendering component goes here
// ...
// Asserting for the rendered mocked data
act( () => {
    (async () => {
        render(null);
        await waitForNextTick();
        expect(container).toMatchSnapshot();
    })()
});
```

I was wondering if anybody found a better solution to this problem?

trojanowski:
@mateusznowosielski it's a known React problem. You can read more about it here: <https://github.com/facebook/react/issues/14769>. There is a WIP PR with a solution: <https://github.com/facebook/react/pull/14853>.

Vpr99:
I just tried @mateusznowosielski's solution and the `expect` call within the async function isn't even being called for me. Is there anything super obvious that I'm doing wrong?

```jsx
it("renders the component", async () => {
  const mockQuery = [
    {
      request: {
        query: gql`
          query HelloQuery {
            hello
          }
        `,
        variables: {},
      },
      result: { data: { hello: "World" } },
    },
  ];
  const waitForNextTick = () => new Promise(resolve => setTimeout(resolve));
  const client = createMockApolloClient(mockQuery);
  const { container, queryByText } = render(
    <ApolloProvider client={client}>
      <MemoryRouter>
        <MyComponent />
      </MemoryRouter>
    </ApolloProvider>
  );
  const loadingStateText = queryByText("Loading!");
  expect(loadingStateText).not.toBeNull();
  act(() => {
    (async () => {
      render(null);
      await waitForNextTick();
      console.log("foo");
      expect(container.firstChild).toMatchSnapshot();
    })();
  });
});llllllllll
```
[...]

PR #103, opened 2019-03-11
Use real `act` in tests

FredyC:
The solution as outlined in https://github.com/trojanowski/react-apollo-hooks/issues/84#issuecomment-470288352

I am using it locally and it works nicely. I suppose that library tests could be updated and drop the mocking. However, given it's all temporary and with async `act` this may not be necessary anymore, I wouldn't really bother :)

### DIFF
--- a/src/internal/actHack.ts
+++ b/src/internal/actHack.ts
@@ -1,8 +1,8 @@
-import { act } from 'react-dom/test-utils'
+import { act } from 'react-dom/test-utils';
 
 export default function actHack(callback: (() => void)) {
   if (process.env.NODE_ENV === 'test') {
-    act(callback)
+    act(callback);
   } else {
     callback();
   }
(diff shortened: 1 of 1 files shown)


FredyC:
@trojanowski Do you see a button to turn this into a real pull request? I've used draft before but suddenly the button is gone.

trojanowski:
@FredyC I can see a "Ready for review" button. But I don't think this change is a good idea - after that, we'd include _react-dom/test-utils_ even in production builds. I prefer to wait for https://github.com/facebook/react/pull/14853 (it should be merged and published soon: https://github.com/facebook/react/issues/14769#issuecomment-470097212) and then to get rid of `actHack`.

FredyC:
Ah, right, haven't thought about that :) Makes sense. 

Weird about that draft thing. They must have changed something or I've messed up :)

FredyC:
I have to use this "hack" for now either way. I don't like `--silent`. I will just use fork for now.

trojanowski:
@FredyC I found an easy way how to get rid of those errors - just use the following snippet:

```typescript
jest.mock('react-apollo-hooks/lib/internal/actHack', () => {
  const { act } = require('react-testing-library');
  return {
    __esModule: true,
    default: act,
  };
});
```

PR #109, opened 2019-03-17
Memoize the function, returned from useMutation

Drago96:
Currently `useMutation` returns a new function reference on each call. This can cause unnecessary re-renders if you are passing the function to child components. This pull request aims to deal with this issue.


Drago96:
Closing due to continuous integration issues with checking typings I have not yet managed to resolve. The issue is weird, because the code compiles and works correctly locally.

rovansteen:
Why close this then? Perhaps someone can help you out. I think this would be a good thing to add.

Issue #117, opened 2019-03-26
data set to {} between re-fetches


FezVrasta:
I created this small example to show my issue:

https://codesandbox.io/s/ovw0m6yl4z

Basically, every time I re-render the component and `useQuery` fetches new data, the `data` object is set to `{}` while `loading` is `true`.

How can I preserve the previously fetched data while I load new data?

Dastari:
I generally store the data in another state variable then update it in a **useEffect** that checks if data exists before assigning the state variable.


```typescript
const MyFuncComponent = () => {
  const [notNulLData, setNotNullData] = useState();
  const { data, error, loading } = useQuery(QUERY);

  useEffect(() => {
    if (data) {
      setNotNullData(data);
    }
  }, [data]);

  return <ComponentThatRendersData data={notNullData} />;
};
```

Not sure if there is a better way or not.

FezVrasta:
@Dastari thanks, but it would really make sense to fix it in this library... I tried to look at the code but couldn't figure out where the problem is.

FezVrasta:
I tried to add `|| result.loading` to https://github.com/trojanowski/react-apollo-hooks/blob/master/src/useQuery.ts#L140

The issue is the `observableQuery` gets invalidated when the query variable change, so the `observableQuery.getLastResult()` returns undefined, because we are asking the `lastResult` to the wrong `observableQuery`...

trojanowski:
@FezVrasta long term it will be solved with a combination of React Suspense and [Concurrent Rendering](https://www.youtube.com/watch?v=ByBPyMBTzM0) (it should even work in the current version, but it's experimental, and I don't recommend to use it in production). About PR #121 - I don't think it's a good idea - I think it's better to be more explicit than to show stale data. I'm also not sure if it works with [fetchMore](https://www.apollographql.com/docs/react/advanced/caching#fetchMore) - please notice that `loading` is set to `true` during loading more data.

FezVrasta:
@trojanowski react-apollo doesn't empties the data object between refetches, this issue exists exactly because the behavior is different (and generated a bug on our app)

FezVrasta:
Please take a look at the same example, but made with `react-apollo`:
https://codesandbox.io/s/ywp7w756vv

lukewlms:
This is broken at present - upgrading from vanilla React Apollo to this breaks our app because the data disappears as soon as the refresh starts.  Added CR comments, please prioritize if possible!

[...]

PR #121, opened 2019-03-28
fix: show previous data while loading new one

FezVrasta:
This PR should fix the issue reported in #117

### DIFF
--- a/src/useQuery.ts
+++ b/src/useQuery.ts
@@ -12,7 +12,7 @@ import ApolloClient, {
   WatchQueryOptions,
 } from 'apollo-client';
 import { DocumentNode } from 'graphql';
-import { useContext, useEffect, useMemo, useState } from 'react';
+import { useContext, useEffect, useMemo, useRef, useState } from 'react';
 import { useApolloClient } from './ApolloContext';
 import { SSRContext } from './internal/SSRContext';
 import actHack from './internal/actHack';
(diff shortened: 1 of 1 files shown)


src/useQuery.ts
```diff
@@ -96,6 +96,8 @@ export function useQuery<
       ? 'cache-first'
       : actualCachePolicy;
 
+  const lastResult = React.useRef();
```
FezVrasta: we use this to be able to reference the **actual** previously returned data, we can't use `observableQuery.getLastResult()` because `observableQuery` may get invalidated and return `null`

src/useQuery.ts
```diff
@@ -96,6 +96,8 @@ export function useQuery<
       ? 'cache-first'
       : actualCachePolicy;
 
+  const lastResult = useRef();
```
FezVrasta: we use this to be able to reference the **actual** previously returned data, we can't use `observableQuery.getLastResult()` because `observableQuery` may get invalidated and return `null`

FezVrasta:
I don't know why I'm getting the two type errors reported by CI, the code uses the same  `(smt || {}).data` few lines above and it's perfectly happy with it.

lukewlms:
Thanks for submitting, this is a breaking change that needs a fix!

I suspect the original {}).data code works because it can never return null (so we never get to the right-hand side of the || and there's always .data).  I.e. the "|| {}" is never called.

It also looks like you need to set a template on your useRef call so the compiler knows that you will be setting null OR the result type.

FezVrasta:
Closing in favour of #134

Issue #125, opened 2019-04-04
Query result changing every rerender


standy:
Hello

There is a problem with useQuery returning new object on every render. Helper functions are not cached, so it leads to deoptimization. 
Take a look on example here: https://codesandbox.io/s/7355xwk4qx

PR fixing this: https://github.com/trojanowski/react-apollo-hooks/pull/126

PR #126, opened 2019-04-04
fix: cache whole response (#125)

standy:
Adding helper functions to cache, so useQuery will not create new object on every render


trojanowski:
Thanks @standy. Published as v0.4.5.

Issue #40, opened 2018-12-20
Merging into react-apollo?


fbartho:
Were you planning on or Would you object to somebody taking this repo and trying to contribute this back to react-apollo?

I love what you've started here, and I want to help!

trojanowski:
Hi @fbartho 

How do you think this repo could be merged with react-apollo? What should be added or changed?

I wasn't very active here recently, but it should improve. I'm going to actively maintain this library and use it in a new project. I'd like the hooks to be added to the official react-apollo, and I'd be happy to help with the implementation. But I also think it would make sense for this library to exist even after that - as something like "react-apollo-lite" - e. g. for new projects where you don't need non-hooks API.

In the next releases (in addition to bugfixes) I'd like to add:

* `useSubscription` (#22 and #37)
* server-side rendering (#8) - also with react-apollo compatibility
* TypeScript rewrite (#39)
* (probably) changes to error management (#28)

I'd appreciate any help. Please let me know what do you think.

fbartho:
Hiya! Based on my readings, if we do our jobs right, places that care about code-file-size can use tree-shaking to shake out the parts they don't use. So… if we do a good job, I'd almost say we "just" need to contribute the hooks bits back into core!

I do agree that the TypeScript-first code is probably a requirement for getting this into core.

Besides that, each of the other features you describe sound useful and nice to have.

Were you thinking about "other" things that you wanted to add to this "react-apollo-lite" that wouldn't be accepted by react-apollo (core)?

trojanowski:
> Were you thinking about "other" things that you wanted to add to this "react-apollo-lite" that wouldn't be accepted by react-apollo (core)?

Not right now. So I agree that if we add hooks to react-apollo and can use just them without adding too much additional stuff to the app bundles, we will depreciate that library and recommend to use the official hooks.

rosskevin:
I'm a comitter on react-apollo (but not an Apollo employee) and have some thoughts. @trojanowski please dm me on the Apollo slack.  I do think there is an opportunity to merge this work there, but things can move quickly here unconstrained. I'm interested to debate pros/cons and perhaps adding a couple contributors here to help move forward until the core team is ready to look at hooks.

JoviDeCroock:
The first thing that would need to be pushed forward in `react-apollo` to make this happen is the new context API.

[...]

PR #135, opened 2019-04-19
[alternative] feat(useMutation): returns tuple with result data

trojanowski:
PR #93 with changes described in https://github.com/trojanowski/react-apollo-hooks/pull/93#issuecomment-478709785 and https://github.com/trojanowski/react-apollo-hooks/pull/93#issuecomment-479462989


lukasluecke:
I believe no one has any concerns against adding the basic functionality / making the BC - maybe we can just get this one merged and discuss the additional changes of #93 afterwards? 🙂

ZVER3D:
Honestly, who cares about all those error boundaries and stuff? Most of the times people would create their own custom hooks based on these ones, depending on their own preferences and projects they are working on.
It would've been enough to just introduce everything from react-apollo as it is into hooks.

SavePointSam:
It appears we are waiting on a response to a comment on the [react-apollo 3.0.0 PR](https://github.com/apollographql/react-apollo/pull/2892#issuecomment-486555822).

> Have you already thought what the exact API for useMutation would look like? We're going to introduce a breaking change here: trojanowski/react-apollo-hooks#93, but I'd prefer the migration to the official hooks to be as painless as possible.

Assuming the API is compatible with the currently proposed API change here, is there any other work to be done here? I would suggest it be in the best interest to keep the API of this package as minimal as possible given that it is already [marked for deprecation in favor of react-apollo 3.0.0](https://github.com/apollographql/react-apollo/pull/2892#issuecomment-486555822).

> The plan is to deprecate react-apollo-hooks after there is official support for hooks in react-apollo (and recommend official hooks instead). There is more info about this in this thread: trojanowski/react-apollo-hooks#40. Until then, I'm going to maintain that library. We're going to release a new version soon: trojanowski/react-apollo-hooks#138.

Given that, the extra (and very interesting) work provided by @FredyC in the original version of this PR (#93), would not be added into this package. Though I also believe those concepts should be explored in the future via either an additive package to react-apollo or explored as an integration to react-apollo.

trojanowski:
> It appears we are waiting on a response to a comment on the react-apollo 3.0.0 PR.

I'm going to add a few more (smaller) breaking changes as described in #138 and release them together. Right now there is an alpha version of react-apollo-hooks available on npm as (`react-apollo-hooks@alpha`), which has this PR merged-in.

> Assuming the API is compatible with the currently proposed API change here, is there any other work to be done here? I would suggest it be in the best interest to keep the API of this package as minimal as possible given that it is already marked for deprecation in favor of react-apollo 3.0.0.

There is no timeline (or at least I wasn't able to find it) when react-apollo 3.0 will be available, so I'm going to release a  stable version with this PR soon (I'd abort if @hwillson or someone else from the Apollo team said the API in react-apollo 3.x would be different).

hwillson:
@trojanowski We don't have a timeline yet, but we're hoping to share a date soon. I'm still optimistically planning on having a Hooks alpha ready by the end of this week (which is looking like Sunday at 11:59 pm at this point). It will be a bit rough around the edges, but it will help kickstart further discussions (and will definitely change over the next few weeks). We're taking the `react-apollo` major version bump as an opportunity to re-architect a few additional (non-Hooks) things, which will likely delay the release of RA 3.0 a bit. As for the `useMutation` API, I think this looks great!

src/useMutation.ts
```diff
@@ -32,18 +36,149 @@ export type MutationFn<TData, TVariables> = (
...
   mutation: DocumentNode,
-  {
-    client: overrideClient,
-    ...baseOptions
-  }: MutationHookOptions<TData, TVariables, TCache> = {}
-): MutationFn<TData, TVariables> {
-  const client = useApolloClient(overrideClient);
-
-  return options => client.mutate({ mutation, ...baseOptions, ...options });
+  baseOptions: MutationHookOptions<TData, TVariables> = {}
```
Cerberus: I suggest to extract `client` from `baseOptions` (like before).

Because `client` must not pass into `objToKey ` on [the line below](https://github.com/trojanowski/react-apollo-hooks/pull/135/files#diff-93bc3bb954808aa3a37a149008cd065fR165)

It might lead a bug when use `useMutation(..., { client: customClient })`

```suggestion
  {	
    client: overrideClient,	
    ...baseOptions,
  }: MutationHookOptions<TData, TVariables, TCache> = {}
): MutationFn<TData, TVariables> {
  const client = useApolloClient(overrideClient);
```

Cerberus: Actually, I tried it and this is what a bug I got when `client` has a circular field inside.

```
Converting circular structure to JSON
    --> starting at object with constructor 'DefaultClient'
    |     property 'localState' -> object with constructor 'LocalState'
```

slackday: I had the same issue. Make sure you convert all 
`const toggleTask = useMutation(TOGGLE_TASK_MUTATION);` 
to 
`const [toggleTask] = useMutation(TOGGLE_TASK_MUTATION);`
It fixed the problem for me

src/useMutation.ts
```diff
@@ -21,6 +24,7 @@ export type MutationUpdaterFn<TData = Record<string, any>> = (
 export interface BaseMutationHookOptions<TData, TVariables>
   extends Omit<MutationOptions<TData, TVariables>, 'mutation' | 'update'> {
   update?: MutationUpdaterFn<TData>;
+  rethrow?: boolean;
```
amiiit: Would it be possible to add callbacks to `onCompleted` and `onError` like the react-apollo's Mutation component?

hiovi:
When will this feature be released? I think it's essential to handling mutations.

trojanowski:
It's already published as a part of version 0.5.0 (which now has the "latest" tag on npm).

src/useMutation.ts
peoplenarthax: > I suggest to extract `client` from `baseOptions` (like before).
> 
> Because `client` must not pass into `objToKey ` on [the line below](https://github.com/trojanowski/react-apollo-hooks/pull/135/files#diff-93bc3bb954808aa3a37a149008cd065fR165)
> 
> It might lead a bug when use `useMutation(..., { client: customClient })`

I get exactly this... only way to bypassing it is by setting the context... Do we know if this is being worked?

PR #141, opened 2019-04-24
fix refetch after error

### LINKED ISSUE #74: useQuery does not return data on refetch after error.
Steps to reproduce:

1. Make a successful request
2. Take graphql server offline
3. Make a new request
4. We get an error now.
5. Take graphql server online  
6. Invoke `refetch`.
7. Note the successful network request in devtools, but we never get data back from useQuery.

sijad:
fixes #74

### DIFF
--- a/src/useQuery.ts
+++ b/src/useQuery.ts
@@ -188,6 +188,8 @@ export function useQuery<
         return;
       }
 
+      let subscription: ZenObservable.Subscription | undefined;
+
       const invalidateCurrentResult = () => {
         // A hack to get rid React warnings during tests. The default
         // implementation of `actHack` just invokes the callback immediately.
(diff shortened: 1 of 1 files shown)


src/useQuery.ts
```diff
@@ -197,16 +199,45 @@ export function useQuery<
...
+
+      // from: https://github.com/apollographql/react-apollo/blob/master/src/Query.tsx#L363
+      // after a error on refetch, without this fix, refetch never works again
+      function invalidateErrorResult() {
+        unsubscribe();
+
+        const lastError = observableQuery.getLastError();
+        const lastResult = observableQuery.getLastResult();
+
+        if (!suspend) {
```
sijad: without this suspend test would fails, I'm not sure if it will breaks something.

SimenB:
@trojanowski any chance of merging and releasing this? 🙏 It fixes the error I'm seeing in our app 🙂

trojanowski:
Could you add a unit test for it? To ensure that it won't be back after #134 is merged.

sijad:
@trojanowski a test case has been added but I think it's a little ugly. the other approach that came to my mind was to expose `refetch` in `Tasks` via `useImperativeHandle`. please let me know how can I make it better

SimenB:
`loading` should not be `true` when refetching, you should check if `networkStatus` is `4`: https://www.apollographql.com/docs/react/api/react-apollo/#datanetworkstatus

jack-sf:
Oh, okay. Well, in that case, the `networkStatus` isn't being updated, when `refetch()` is done after an error occurs.

See on that screenshot:

image

It goes:
- loading
- failed
- ready

Instead, it probably should be:
- loading
- failed
- refetch (or loading?)
- ready

amannn:
Any update on this PR? Would be really great to see this merged.

sijad:
you should probably use [@apollo/react-hooks](https://www.npmjs.com/package/@apollo/react-hooks)

huidini:
> 
> 
> you should probably use [@apollo/react-hooks](https://www.npmjs.com/package/@apollo/react-hooks)

I started using it as well, but I really really miss the elegance and ease of use of Suspense.

amannn:
I understand. However `@apollo/react-hooks` is in beta currently and I wouldn't want to move to that library before it's stable.

Is there something missing in this PR or could it be merged as-is? As @huidini mentioned, using Suspense might still be a feature only this library has.

sijad:
@amannn `@apollo/react-hooks` 3.0.0 released about an hour ago

https://github.com/apollographql/react-apollo/releases/tag/%40apollo%2Freact-hooks%403.0.0

amannn:
@sijad Oh wow, what a coincidence 🙂. Thanks!

Issue #127, opened 2019-04-08
Typescript complains that data prop in query hook can be undefined 


sami616:
This makes destructuring the data prop very hard.

```
  const {
    data: { user }, // because data prop doesn't fallback to an empty object this destructuring fails  
    loading,
    error,
  } = useQuery<Todos.Data>(TODOS)
```

Looks like a fix for this want merged into ```react-apollo```

https://github.com/apollographql/react-apollo/pull/1983

warreee:
I'm having the same issue.
An empty object would indeed be nicer.

trojanowski:
`data` is undefined when `skip` option is equal to `true`. It's similar how react-apollo works: <https://github.com/apollographql/react-apollo/blob/da452d117bdad8da6a87ada23c793b2eeb0fea7c/src/Query.tsx#L64>.

warreee:
@trojanowski thanks for the reply, indeed it seems to be the same.
However, I haven't found out why react-apollo returns at least an empty object and this library returns either TData or undefined:
`const queryResult = useQuery<TData, TVariables>(query);`
Any idea?

PR #146, opened 2019-04-29
Fix fallback of `data` attribute in query result

sgoll:
Related to #127.

As described in https://www.apollographql.com/docs/react/essentials/queries#props, the `data` object is either the actual query result, or the empty object `{}` when the data is not (yet) available.

The type of `QueryHookResult` should match that.


sgoll:
Alright, it seems it's more complicated than this. I guess we'd need a type union to differentiate between loading and/or failed queries (that have `data` of `{}`) and completed queries (that have actual `data`).

And then there's partial queries also …

PR #159, opened 2019-05-09
Add partialRefetch as an option

### LINKED ISSUE #128: partialRefetch as an option
The partialRefetch option is not available at the moment. Would it be possible to make it available?

More information here: https://www.apollographql.com/docs/react/essentials/queries

estrada9166:
fix #128 

Taken from [here](https://github.com/apollographql/react-apollo/blob/22f8ebf52b26b348d6be905d5b7fbbfea51c1541/src/Query.tsx#L455-L472)

### DIFF
--- a/src/useQuery.ts
+++ b/src/useQuery.ts
@@ -172,12 +172,14 @@ export function useQuery<
         result.partial &&
         fetchPolicy !== 'cache-only'
       ) {
-        Object.assign(result.data, helpers, {
+        const partialRefetchResult = {
+          ...result.data,
+          ...helpers,
           loading: true,
           networkStatus: NetworkStatus.loading,
-        });
+        };
         helpers.refetch();
-        return result.data;
+        return partialRefetchResult;
       }
 
       return {
--- a/src/useQuery.ts
+++ b/src/useQuery.ts
@@ -42,6 +42,7 @@ export interface QueryHookOptions<TVariables, TCache = object>
   ssr?: boolean;
   skip?: boolean;
   suspend?: boolean;
+  partialRefetch?: boolean;
 }
 
 export interface QueryHookResult<TData, TVariables>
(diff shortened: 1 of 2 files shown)


trojanowski:
Thanks @estrada9166. Could you add a unit test for it?

src/useQuery.ts
```diff
@@ -163,6 +165,23 @@ export function useQuery<
...
 
+      // Taken from https://github.com/apollographql/react-apollo/blob/22f8ebf52b26b348d6be905d5b7fbbfea51c1541/src/Query.tsx#L455-L472
+      if (
+        partialRefetch &&
+        Object.keys(result.data).length === 0 &&
+        result.partial &&
+        fetchPolicy !== 'cache-only'
+      ) {
+        const partialRefetchResult = {
+          ...result.data,
```
brabeji: These are empty so maybe it would be better, to keep stale data from previous result using `useRef`.

brabeji: or maybe handle this here in a similar way as error is https://github.com/trojanowski/react-apollo-hooks/pull/159/files#diff-9affe4a75e3f039c2a00e27c8bf815aeR148

brabeji: BTW, sorry to invade this PR without introduction. I didn't catch it in time and coded basically the same thing yesterday evening. So now I'm just sharing my experience. Thanks for the PR!

estrada9166: Thanks for the comments! I'll update the PR with this! is much better.

brabeji: FWIW Also this one is related to the need of showing stale data:
https://github.com/trojanowski/react-apollo-hooks/pull/134

trufa:
@brabeji Any new for this? :) Thanks!

brabeji:
@trufa I'm using a locally patched version of `react-apollo-hooks` (via `patch-package`) and waiting for official support from `react-apollo`: https://github.com/apollographql/react-apollo/pull/2892#issuecomment-504544217

PR #169, opened 2019-05-30
Allow `cache-and-network` in apollo-client@^2.6.0

sgoll:
This fixes #163 and is identical to PR #165 except that it also upgrades the `devDependencies` to the required versions of apollo-client and typescript.

PR #181, opened 2019-06-18
Update useQuery to utilize WatchQueryOptions

jamesreggio:
It's more accurate this way, and a recent change has removed `cache-and-network` from the `fetchPolicy` values allowed in a plain old `QueryOptions` object, meaning that without this change, it's not possible to use `cache-and-network` with the `useQuery` hook.

Related change in Apollo Client: https://github.com/apollographql/apollo-client/commit/cf069bc7ee6577092234b0eb0ac32e05d50f5a1c#diff-88b34d7add139a9c0e7d4c2b1944a8dcR18

### DIFF
--- a/src/useQuery.ts
+++ b/src/useQuery.ts
@@ -8,7 +8,6 @@ import ApolloClient, {
   NetworkStatus,
   ObservableQuery,
   OperationVariables,
-  QueryOptions,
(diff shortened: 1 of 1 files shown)


jamesreggio:
Derp. Duplicate of #169.

